PR #9, opened 2023-06-20
refactor: Common static analysis issues

sambsnyd:
Use this link to re-run the recipe: https://app.moderne.io/recipes/org.openrewrite.staticanalysis.CommonStaticAnalysis?organizationId=T3BlblJld3JpdGU%3D


timtebeek:
Doesn't help readability here; we can revisit this later once this has crystallized a bit more.

PR #10, opened 2023-08-10
Add or remove imports

timtebeek:


src/main/java/org/openrewrite/java/template/RefasterTemplateProcessor.java
```diff
@@ -257,6 +257,17 @@ public void visitClassDef(JCTree.JCClassDecl classDecl) {
...
+
+                        for (String import_ : imports) {
+                            recipe.append("                maybeRemoveImport(\"" + import_ + "\");\n");
+                            recipe.append("                maybeAddImport(\"" + import_ + "\");\n");
+                        }
+                        for (String import_ : staticImports) {
+                            recipe.append("                maybeRemoveImport(\"" + import_ + "\");\n");
+                            int dot = import_.lastIndexOf('.');
+                            recipe.append("                maybeAddImport(\"" + import_.substring(0, dot) + "\", \"" + import_.substring(dot + 1) + "\");\n");
+                        }
```
timtebeek, line 269: I was surprised we don't seem to separate imports, even though they end up OK into each of the before/after templates. That meant here I've had to both add and remove them _maybe_; which was good enough to get working, but not something that I'd be confident relying on.

timtebeek:
Do we see any reasonable ways to unit test this stuff here? Not looking to slow us down, but it'd be slightly nicer than having to publish to maven local.

timtebeek:
Also quickly tried out testing in https://github.com/openrewrite/rewrite-templating/pull/11

knutwannheden:
I would prefer if we could try to instead apply `RemoveUnusedImports` and `ShortenFullyQualifiedTypeReferences`. That would then also allow us to use fully qualified names everywhere in the template code and not require the template to be aware of any imports. That in turn has the benefit that there won't be any collisions when applying the template to some compilation unit, which may have a colliding import declared already. I could look into this, if it is urgent. But I think it should actually be rather simple (a matter of generating two `doAfterVisit()` calls), so if you want to give it a try...

timtebeek:
Didn't get to a proper fix here, but we do have the tests merged into this one such that we should be able to iterate more easily.
Or we can merge this as is, and work on a proper solution next week, to unblock @AlekSimpson .

knutwannheden (approved):
Let's merge it. I will look more thoroughly at it next week.

timtebeek:
> Let's merge it. I will look more thoroughly at it next week.

Appreciate it! I've added a test to quickly capture the current behavior just now in https://github.com/openrewrite/rewrite-templating/commit/00733505a8ba4dd8348154ca04cc4e707352c14e.

PR #11, opened 2023-08-10
Test annotation processor using compile-testing

timtebeek:
This will need some more work to finish it up, but thought to put the first bit of effort in.


build.gradle.kts
```diff
@@ -65,6 +65,20 @@ val tools = compiler.get().metadata.installationPath.file("lib/tools.jar")
 dependencies {
     compileOnly(files(tools))
     compileOnly("org.jetbrains:annotations:24.0.+")
+
+    // Needed for annotation processing tests
+    testImplementation(files(tools))
+    testImplementation("org.openrewrite:rewrite-java:latest.release")
```
knutwannheden, line 71: I wonder if we should use `latest.integration` instead?

build.gradle.kts
```diff
@@ -87,6 +101,7 @@ configure<LicenseExtension> {
     header = project.rootProject.file("gradle/licenseHeader.txt")
     mapping(kotlin.collections.mapOf("kt" to "SLASHSTAR_STYLE", "java" to "SLASHSTAR_STYLE"))
     strictCheck = true
+    exclude("**/UseStringIsEmpty*.java")
```
knutwannheden, line 104: Should we move the files into a specific `src/test/resources` subdirectory and then exclude all of the files in that subdirectory?

build.gradle.kts
timtebeek: Done! Didn't think it would be necessary often, but best to be consistent I guess. We'll see if nebula picks things up correctly on the next release.

build.gradle.kts
timtebeek: Good suggestion; moved those into `src/test/resources/recipes` and updated the license header exclude.

knutwannheden (approved):
👍

PR #14, opened 2023-08-11
Only add and remove imports as required

knutwannheden:


src/main/java/org/openrewrite/java/template/RefasterTemplateProcessor.java
```diff
@@ -187,24 +187,34 @@ public void visitClassDef(JCTree.JCClassDecl classDecl) {
...
-                            staticImports.add(anImport.owner.getQualifiedName().toString().replace('$', '.') + '.' + anImport.flatName().toString());
+                            staticImports.computeIfAbsent(descriptor.afterTemplate, k -> new TreeSet<>())
+                                    .add(anImport.owner.getQualifiedName().toString().replace('$', '.') + '.' + anImport.flatName().toString());
                         } else {
                             throw new AssertionError(anImport.getClass());
                         }
                     }
 
+                    for (Set<String> imports : imports.values()) {
+                        imports.removeIf(i -> "java.lang".equals(i.substring(0, i.lastIndexOf('.'))));
```
timtebeek, line 213: This is better indeed; the exclude of `java.lang.**` we had before was too broad. Thanks!

timtebeek (approved):
Thanks for doing this! Like how I can now look at the tests to better understand the changes.

PR #16, opened 2023-08-11
Allow the after template to dereference parameters multiple times

knutwannheden:


knutwannheden:
@timtebeek Not very thoroughly tested, but will allow Alek to iterate on this and do some more testing.

PR #17, opened 2023-08-13
Have outer classes extend Recipe too

timtebeek:
Simplifies calling related recipes all together without Yaml recipe lists.

I don't immediately see any possible negative effects: we already generate the class, and having a way to run all of them together should simplify things for the majority of cases.

### DIFF
--- a/src/test/resources/recipes/MultipleDereferencesRecipes.java
+++ b/src/test/resources/recipes/MultipleDereferencesRecipes.java
@@ -8,7 +8,29 @@ import org.openrewrite.java.JavaVisitor;
 import org.openrewrite.java.template.Primitive;
 import org.openrewrite.java.tree.*;
 
-public final class MultipleDereferencesRecipes {
+import java.util.Arrays;
+import java.util.List;
+
+public final class MultipleDereferencesRecipes extends Recipe {
+
+    @Override
+    public String getDisplayName() {
+        return "Refaster recipes for `foo.MultipleDereferences`";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Refaster template recipes for `foo.MultipleDereferences`.";
+    }
+
+
+    @Override
+    public List<Recipe> getRecipeList() {
+        return Arrays.asList(
+                new StringIsEmptyRecipe(),
+                new EqualsItselfRecipe()
+        );
+    }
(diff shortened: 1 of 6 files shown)


knutwannheden:
Possibly wr should actually declare the nested classes with the same visibility as in the Refaster source. That way a nested class can also be private and as a result only the outer recipe would be callable. I think that could be useful.

knutwannheden (approved):

timtebeek:
> Possibly wr should actually declare the nested classes with the same visibility as in the Refaster source. That way a nested class can also be private and as a result only the outer recipe would be callable. I think that could be useful.

Indeed a neat idea to match input and output modifiers for nested classes. Applied a first such change in https://github.com/openrewrite/rewrite-templating/pull/17/commits/8b26e4c3726fea13e8f0fe38d9f2b566c225e66e ; Not done for outer classes, as that would hide the inner ones as well I believe.

src/test/resources/recipes/MultipleDereferencesRecipes.java
```diff
@@ -8,8 +8,30 @@
...
+public final class MultipleDereferencesRecipes extends Recipe {
+
+    @Override
+    public String getDisplayName() {
+        return "Refaster recipes for `foo.MultipleDereferences`";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Refaster template recipes for `foo.MultipleDereferences`.";
```
knutwannheden, line 23: Should we possibly also here include the source code as Markdown?

timtebeek: I'm afraid that would result in a very long blob of text for some of the recipes I anticipate here. Take for instance [the Plexus `StringUtils` methods](https://github.com/codehaus-plexus/plexus-utils/blob/master/src/main/java/org/codehaus/plexus/util/StringUtils.java) we hope to replace with recipes; that's a _long_ list that I think would best be documented on each of the individual recipes.

I do understand there's some tension with whether that recipe would have visible individual recipes, but since we also plan to add a `@DocumentExample` to the docs, then I think including the recipe source here could be too much.

knutwannheden: OK. Let's keep it like this for now. We may want to add an annotation with which we can control the generation of the display name and description.

PR #18, opened 2023-08-13
Call ShortenFullyQualifiedTypeReferences after visit

timtebeek:
## What's changed?
We now call `doAfterVisit(new ShortenFullyQualifiedTypeReferences().getVisitor())` from within the if statement.

## What's your motivation?
I _think_ this is what we discussed, such that templates can use fully qualified references for convenience, yet insert shortened references in practice.

Did you indeed intend to call the visitor like this, or in a more targeted way?

@knutwannheden 

Not sure if there's any other ways we might want to explore.

## Any additional context
This ought to help PRs such as this on
- https://github.com/openrewrite/rewrite-migrate-java/pull/272

### DIFF
--- a/src/main/java/org/openrewrite/java/template/RefasterTemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/RefasterTemplateProcessor.java
@@ -314,6 +314,7 @@ public class RefasterTemplateProcessor extends AbstractProcessor {
                                 recipe.append("                    maybeAddImport(\"" + import_.substring(0, dot) + "\", \"" + import_.substring(dot + 1) + "\");\n");
                             }
                         }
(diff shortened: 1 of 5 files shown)


knutwannheden (commented):
I think adding the `ShortenFullyQualifiedTypeReferences` recipe on its own is not very interesting. We should IMHO change the template generation code so that it only produces fully qualified references to types and static members. But this will be more work, as I think I remember that the Refaster processor currently lets the other annotation processor generate the template code. Possibly we should for more flexibility instead inline it into the Refaster processor.

timtebeek:
As discussed over on https://github.com/openrewrite/rewrite-migrate-java/pull/272#issuecomment-1677321379

PR #19, opened 2023-08-14
Generate classpath entries for FQN references

timtebeek:
## What's changed?
Right now fully qualified references do not generate classpath entries on the JavaTemplate classes, since we only loop over imports to generate those. We expand that to include any type use of classes that need imports.

## What's your motivation?
When we want to [swap between similar classes](https://github.com/openrewrite/rewrite-migrate-java/pull/272) we would want to use FQNs in the recipes for readability.

## Any additional context
- https://github.com/openrewrite/rewrite-migrate-java/pull/272
- https://github.com/openrewrite/rewrite-templating/pull/18

### DIFF
--- a/src/main/java/org/openrewrite/java/template/TemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/TemplateProcessor.java
@@ -27,6 +27,7 @@ import com.sun.tools.javac.tree.JCTree;
 import com.sun.tools.javac.tree.JCTree.JCCompilationUnit;
 import com.sun.tools.javac.tree.TreeScanner;
 import com.sun.tools.javac.util.Context;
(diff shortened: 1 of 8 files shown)


knutwannheden (approved):
I think I saw that this also covers annotations, which is good. I suspect that we will at some point also have to traverse the indirectly referenced types (supertypes, return types, parameter types, etc.), but let's leave it at this for now.

timtebeek:
> I think I saw that this also covers annotations, which is good. I suspect that we will at some point also have to traverse the indirectly referenced types (supertypes, return types, parameter types, etc.), but let's leave it at this for now.

Thanks for the quick review! Indeed I think we can iterate on this still, but since it's functional we can continue the work in those linked PRs and see what else pops up.

PR #20, opened 2023-08-14
Add and remove imports for qualified classes

timtebeek:
## What's changed?
Also add/remove imports for qualified classes in input or output.

## What's your motivation?
Qualified classes in before/after did not get their imports added/removed as necessary.


timtebeek:
An initial change in the tests to demonstrate the issue, before a followup that ought to add / remove those imports again.

src/test/resources/recipes/ShouldAddImports.java
```diff
@@ -15,14 +13,14 @@ String before(String s) {
...
-            return Objects.toString(s);
+            return java.util.Objects.toString(s);
         }
     }
 
     public static class ObjectsEquals {
         @BeforeTemplate
         boolean before(int a, int b) {
-            return Objects.equals(a, b);
+            return java.util.Objects.equals(a, b);
```
timtebeek, line 23: @knutwannheden with this change the `RefasterTemplateProcessor` no longer generates the `maybeAdd/RemoveImport` lines for any FQN added or removed. The workaround is simple: don't fully qualify types in the input, but that's a bit unfortunate.

I'm wondering if makes sense to have a type detector, next to the import and classpath detectors we already have, and determine the types used in input and output. When you then have two sets of types, you can determine which imports to maybe add / remove from the visitor. Not sure if we would then still need the import detector, or if there are any side effects I'm missing right now. What are your thoughts here?

knutwannheden:
Is this PR still relevant with respect to our discussion to always generate FQNs into the template code?

timtebeek:
Closing as not something to prioritize now. Should be rare that we need FQN in before or after templates.

PR #22, opened 2023-08-17
Add `doAfterVisit(UnnecessaryParentheses)` to generated recipes

timtebeek:
## What's your motivation?
Until we update the JavaTemplate to detect and insert parentheses around for instance binary expressions, we need a quick way to move along https://github.com/openrewrite/rewrite-migrate-java/pull/271 ; With this change we're able to add parentheses to the templates defined there, which are then cleaned up through `UnnecessaryParentheses`.

We could _always_ add parentheses in the JavaTemplate Strings produced by `TemplateProcessor` here; that might be a bit excessive, but would prevent us from having to add those parentheses to rewrite-migrate-java.

## Any additional context
- https://github.com/openrewrite/rewrite-migrate-java/pull/271

### DIFF
--- a/src/test/resources/recipes/MultipleDereferencesRecipes.java
+++ b/src/test/resources/recipes/MultipleDereferencesRecipes.java
@@ -54,7 +54,7 @@ public final class MultipleDereferencesRecipes extends Recipe {
                     JavaTemplate.Matcher matcher;
                     if ((matcher = before.matcher(getCursor())).find()) {
                         doAfterVisit(new org.openrewrite.java.ShortenFullyQualifiedTypeReferences().getVisitor());
-                        doAfterVisit(new org.openrewrite.staticanalysis.UnnecessaryParentheses().getVisitor());
+                        doAfterVisit(new org.openrewrite.java.cleanup.UnnecessaryParenthesesVisitor());
                         return after.apply(getCursor(), elem.getCoordinates().replace(), matcher.parameter(0), matcher.parameter(0));
                     }
                     return super.visitMethodInvocation(elem, ctx);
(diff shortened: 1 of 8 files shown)


knutwannheden:
Another alternative would be for us to determine whether expressions are required or not at "insertion time"; i.e. we could add some kind of `maybeParenthesize(Expression, Cursor)` utility method. I imagine this could be useful in other recipes as well. But generating a `doAfterVisit()` is simpler of course.

My only worry with `doAfterVisit()` is that if a lot of modifications are made, then for each such modification we schedule one of these visitors. I wonder if we should either add an overload or another method (like `doOnceAfterVisit()`) or add a marker interface to visitors. For `maybeAddImport()` and `maybeRemoveImport()` we hand-coded this logic.

src/test/resources/recipes/MultipleDereferencesRecipes.java
```diff
@@ -54,7 +53,8 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
                 public J visitMethodInvocation(J.MethodInvocation elem, ExecutionContext ctx) {
                     JavaTemplate.Matcher matcher;
                     if ((matcher = before.matcher(getCursor())).find()) {
-                        doAfterVisit(new ShortenFullyQualifiedTypeReferences().getVisitor());
+                        doAfterVisit(new org.openrewrite.java.ShortenFullyQualifiedTypeReferences().getVisitor());
+                        doAfterVisit(new org.openrewrite.staticanalysis.UnnecessaryParentheses().getVisitor());
```
knutwannheden, line 57: Actually, we have `UnnecessaryParenthesesVisitor` in `rewrite-java`: `org.openrewrite.java.cleanup.UnnecessaryParenthesesVisitor`.

knutwannheden (commented):
Left one small comment. After that I am happy.

PR #24, opened 2023-08-17
Generate uses type and uses method

### LINKED ISSUE #23: Generate `UsesType` and `UsesMethod` to limit recipe execution and improve performance
`UsesType` might be fairly easy given that we already detect imports.

`UsesMethod` would require us to extract those separately, would reduce the need for `UsesType` I suppose.

timtebeek:
## What's changed?
Generate Preconditions based on imports and static imports used in before templates.

## What's your motivation?
Speed up recipe execution.
Fixes #23

Would we want to detect additional methods now already?
Figured this was good enough to start to limit recipe execution.

We could add a new detector specific for methods used in the before templates, but that's perhaps more work we can do later.

## Any additional context
- https://github.com/openrewrite/rewrite-templating/issues/23
- https://github.com/openrewrite/rewrite-templating/pull/22/
- https://github.com/openrewrite/rewrite-migrate-java/pull/271

### DIFF
--- a/src/main/java/org/openrewrite/java/template/RefasterTemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/RefasterTemplateProcessor.java
@@ -288,14 +288,15 @@ public class RefasterTemplateProcessor extends AbstractProcessor {
                                 .collect(Collectors.toSet());
                         Set<String> usedTypes = new LinkedHashSet<>();
                         for (String import_ : imports.values().stream().flatMap(Set::stream).collect(Collectors.toSet())) {
+                            if (beforeImports.contains(import_)) {
+                                usedTypes.add(import_);
+                            }
                             if (import_.startsWith("java.lang.")) {
                                 continue;
                             }
                             if (beforeImports.contains(import_) && afterImports.contains(import_)) {
-                                usedTypes.add(import_);
                             } else if (beforeImports.contains(import_)) {
                                 recipe.append("                    maybeRemoveImport(\"" + import_ + "\");\n");
-                                usedTypes.add(import_);
                             } else if (afterImports.contains(import_)) {
                                 recipe.append("                    maybeAddImport(\"" + import_ + "\");\n");
                             }
@@ -317,15 +318,16 @@ public class RefasterTemplateProcessor extends AbstractProcessor {
                                 .collect(Collectors.toSet());
                         Set<String> usedMethods = new LinkedHashSet<>();
                         for (String import_ : staticImports.values().stream().flatMap(Set::stream).collect(Collectors.toSet())) {
+                            int dot = import_.lastIndexOf('.');
+                            if (beforeImports.contains(import_)) {
+                                usedMethods.add(import_.substring(0, dot) + ' ' + import_.substring(dot + 1) + "(..)");
+                            }
                             if (import_.startsWith("java.lang.")) {
                                 continue;
                             }
-                            int dot = import_.lastIndexOf('.');
                             if (beforeImports.contains(import_) && afterImports.contains(import_)) {
-                                usedMethods.add(import_.substring(0, dot) + ' ' + import_.substring(dot + 1) + "(..)");
                             } else if (beforeImports.contains(import_)) {
                                 recipe.append("                    maybeRemoveImport(\"" + import_ + "\");\n");
-                                usedMethods.add(import_.substring(0, dot) + ' ' + import_.substring(dot + 1) + "(..)");
                             } else if (afterImports.contains(import_)) {
                                 recipe.append("                    maybeAddImport(\"" + import_.substring(0, dot) + "\", \"" + import_.substring(dot + 1) + "\");\n");
                             }
--- a/src/main/java/org/openrewrite/java/template/RefasterTemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/RefasterTemplateProcessor.java
@@ -447,7 +447,7 @@ public class RefasterTemplateProcessor extends AbstractProcessor {
 
                     Set<String> localImports = imports.getOrDefault(beforeTemplate, Collections.emptySet());
                     for (String import_ : localImports) {
-                        usesVisitors.add("new UsesType<>(\"" + import_ + "\", false)");
+                        usesVisitors.add("new UsesType<>(\"" + import_ + "\", true)");
                     }
                     Set<String> localStaticImports = staticImports.getOrDefault(beforeTemplate, Collections.emptySet());
                     for (String import_ : localStaticImports) {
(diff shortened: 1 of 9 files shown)


src/main/java/org/openrewrite/java/template/RefasterTemplateProcessor.java
```diff
@@ -301,18 +310,26 @@ public void visitClassDef(JCTree.JCClassDecl classDecl) {
                                 .map(Map.Entry::getValue)
                                 .flatMap(Set::stream)
                                 .collect(Collectors.toSet());
+                        Set<String> usedMethods = new LinkedHashSet<>();
                         for (String import_ : staticImports.values().stream().flatMap(Set::stream).collect(Collectors.toSet())) {
                             if (import_.startsWith("java.lang.")) {
                                 continue;
```
knutwannheden, line 316: We miss out on some used methods here.

src/main/java/org/openrewrite/java/template/RefasterTemplateProcessor.java
```diff
@@ -280,17 +281,25 @@ public void visitClassDef(JCTree.JCClassDecl classDecl) {
...
                                 recipe.append("                    maybeRemoveImport(\"" + import_ + "\");\n");
+                                usedTypes.add(import_);
                             } else if (afterImports.contains(import_)) {
                                 recipe.append("                    maybeAddImport(\"" + import_ + "\");\n");
                             }
                         }
+                        if (usedTypes.size() == 1) {
+                            preconditions.add("new UsesType<>(\"" + usedTypes.iterator().next() + "\", false)");
+                        } else if (1 < usedTypes.size()) {
+                            preconditions.add("Preconditions.and(" + usedTypes.stream().map(t -> "new UsesType<>(\"" + t + "\", false)").collect(Collectors.joining(", ")) + ')');
```
knutwannheden, line 301: Will this `and()` work correctly in the presence of multiple before templates? If a type is used by all before templates we can `and` it, others will probably have to be `or`ed. So we would have to build up a tree, basically.

src/main/java/org/openrewrite/java/template/RefasterTemplateProcessor.java
timtebeek: Good call; picked up in https://github.com/openrewrite/rewrite-templating/pull/24/commits/34e9ba9df053ff2ef5b81b8e9b35f90bb06da8ec

src/main/java/org/openrewrite/java/template/RefasterTemplateProcessor.java
timtebeek: It should; we build up an `and()` precondition per before template, looking only at imports present in the specific before template. The [further down](https://github.com/openrewrite/rewrite-templating/pull/24/files#diff-51aa8b78b7dec3df0eb18b0aee93bba1f4ea79f3b797ddddaa116cbb0bea5318R360-R364) we compose those together into a `or()` precondition, using each of the before template `and()` precondition.

At both stages I took care to generate the minimal `Precondition` we could get away with, but I started out testing with  a more nested approach by default in 37a2b8c5e8e91e92614d4d2616550f574c1c9134, which is still generated when needed.

src/test/resources/recipes/ShouldAddImportsRecipes.java
```diff
@@ -100,7 +106,47 @@ public J visitMethodInvocation(J.MethodInvocation elem, ExecutionContext ctx) {
                 }
 
             };
+            return Preconditions.check(
+                    new UsesType<>("java.util.Objects", false),
```
knutwannheden, line 110: Out of curiosity: Why is there no `UsesMethod` for `Objects#equals()` here? I would have expected the precondition to check for either `Objects#equals()` or `Integer#compare()`.

timtebeek: Currently those would only be generated if we use a static import of those methods; we can expand that of course, but figured this first implementation already limits the execution of the recipes to recipes using the type, which at scale already trims it down quite a bit.

knutwannheden: Thanks for the explanation. Yes, also covering the others would indeed make sense.

src/test/resources/recipes/NestedPreconditionsRecipe.java
```diff
@@ -0,0 +1,59 @@
...
+                    doAfterVisit(new org.openrewrite.java.cleanup.UnnecessaryParenthesesVisitor());
+                    return hashtable.apply(getCursor(), elem.getCoordinates().replace(), matcher.parameter(0));
+                }
+                return super.visitExpression(elem, ctx);
+            }
+
+        };
+        return Preconditions.check(
+                Preconditions.or(
+                        Preconditions.and(new UsesType<>("java.util.HashMap", false), new UsesType<>("java.util.Map", false)),
```
knutwannheden, line 55: Since `Map` is not necessarily directly referenced by the input source, I think we should either leave it out or use the overload with the `includeImplicit` parameter. Otherwise the precondition may be too strict. For example:
```java
Collection m = new HashMap();
```

timtebeek: Hmm; we don't differentiate between in and output types when we detect imports right now. Would passing `true` to `usesType` https://github.com/openrewrite/rewrite-templating/pull/24/commits/e7d0883eca9e2adb54de36f20e607818a0997969 be enough to reduce your concern here?

src/test/resources/recipes/ShouldAddImportsRecipes.java
timtebeek: Logged as an improvement:
- https://github.com/openrewrite/rewrite-templating/issues/26

src/test/resources/recipes/NestedPreconditionsRecipe.java
```diff
@@ -0,0 +1,59 @@
...
+                    doAfterVisit(new org.openrewrite.java.cleanup.UnnecessaryParenthesesVisitor());
+                    return hashtable.apply(getCursor(), elem.getCoordinates().replace(), matcher.parameter(0));
+                }
+                return super.visitExpression(elem, ctx);
+            }
+
+        };
+        return Preconditions.check(
+                Preconditions.or(
+                        Preconditions.and(new UsesType<>("java.util.HashMap", true), new UsesType<>("java.util.Map", true)),
```
knutwannheden, line 55: For HashMap we could actually pass the second argument as `true`, because it must be directly referenced. `Map` is different, because you got that from the template's return type and not the template body. But we can still fix this later.

knutwannheden (approved):

src/test/resources/recipes/NestedPreconditionsRecipe.java
timtebeek: Thanks! Captures as a note here: https://github.com/openrewrite/rewrite-templating/issues/26#issuecomment-1683704956

Issue #28, opened 2023-08-18
Simplify boolean expressions after a replacement


timtebeek:
## What problem are you trying to solve?
A lot of after templates in https://github.com/openrewrite/rewrite-migrate-java/pull/271 produce boolean expressions to handle nullability. In some cases, the boolean expressions don't make sense if the argument is a literal, such as with
```java
class EqualsIgnoreCase {
    @BeforeTemplate
    boolean before(String s, String other) {
        return StringUtils.equalsIgnoreCase(s, other);
    }
    @AfterTemplate
    boolean after(String s, String other) {
        return (s == null && other == null || s != null && s.equalsIgnoreCase(other));
    }
}
```
Which replaces the following
```diff
- boolean bool = StringUtils.equalsIgnoreCase(in, "other");
+ boolean bool = in == null && "other" == null || in != null && in.equalsIgnoreCase("other");
```

## Describe the solution you'd like
The ` && "other" == null` can likely be dropped through [SimplifyBooleanExpression](https://docs.openrewrite.org/recipes/staticanalysis/simplifybooleanexpression) applied in `doAfterVisit` [where applicable](https://github.com/openrewrite/rewrite-templating/issues/25).
```diff
- boolean bool = StringUtils.equalsIgnoreCase(in, "other");
+ boolean bool = in == null || in != null && in.equalsIgnoreCase("other");
```

## Have you considered any alternatives or workarounds?
No alternatives here I think, but a requirement for acceptance of the replaced suggestions.

## Additional context
- https://github.com/openrewrite/rewrite-migrate-java/pull/271

PR #31, opened 2023-08-18
Clean up nullable and empty string comparisons after visit

timtebeek:
## What's changed?
We call `SimplifyBooleanExpression` after visit.

## What's your motivation?
Replacements such as this one take two arguments, and have to be null safe depending on those arguments.
```java
private static class EqualsIgnoreCase {
    @BeforeTemplate
    boolean before(String s, String other) {
        return StringUtils.equalsIgnoreCase(s, other);
    }

    @AfterTemplate
    boolean after(String s, String other) {
        return (s == null && other == null || s != null && s.equalsIgnoreCase(other));
    }
}
```
When the above is passed a literal as second argument, it looks odd to do `"other" == null`.
With this do after visit we simplify those expressions again for easier recipe definition and better end result.

Fixes https://github.com/openrewrite/rewrite-templating/issues/28

## Any additional context
Requires
- [ ] https://github.com/openrewrite/rewrite-static-analysis/pull/154

- https://github.com/openrewrite/rewrite-templating/issues/28
- https://github.com/openrewrite/rewrite-templating/pull/29

### DIFF
--- a/src/main/java/org/openrewrite/java/template/RefasterTemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/RefasterTemplateProcessor.java
@@ -94,7 +94,8 @@ public class RefasterTemplateProcessor extends AbstractProcessor {
 
     static Set<String> DO_AFTER_VISIT = Stream.of(
             "new org.openrewrite.java.ShortenFullyQualifiedTypeReferences().getVisitor()",
-            "new org.openrewrite.java.cleanup.UnnecessaryParenthesesVisitor()"
+            "new org.openrewrite.java.cleanup.UnnecessaryParenthesesVisitor()",
+            "new org.openrewrite.staticanalysis.SimplifyBooleanExpression().getVisitor()"
     ).collect(Collectors.toCollection(LinkedHashSet::new));
 
     static ClassValue<List<String>> LST_TYPE_MAP = new ClassValue<List<String>>() {
--- a/src/main/java/org/openrewrite/java/template/RefasterTemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/RefasterTemplateProcessor.java
@@ -95,7 +95,7 @@ public class RefasterTemplateProcessor extends AbstractProcessor {
     static Set<String> DO_AFTER_VISIT = Stream.of(
             "new org.openrewrite.java.ShortenFullyQualifiedTypeReferences().getVisitor()",
             "new org.openrewrite.java.cleanup.UnnecessaryParenthesesVisitor()",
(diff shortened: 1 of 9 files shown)


src/main/java/org/openrewrite/java/template/RefasterTemplateProcessor.java
```diff
@@ -94,7 +94,8 @@ public class RefasterTemplateProcessor extends AbstractProcessor {
 
     static Set<String> DO_AFTER_VISIT = Stream.of(
             "new org.openrewrite.java.ShortenFullyQualifiedTypeReferences().getVisitor()",
-            "new org.openrewrite.java.cleanup.UnnecessaryParenthesesVisitor()"
+            "new org.openrewrite.java.cleanup.UnnecessaryParenthesesVisitor()",
+            "new org.openrewrite.staticanalysis.SimplifyBooleanExpression().getVisitor()"
```
knutwannheden, line 98: This visitor also still exists in `rewrite-java` as `org.openrewrite.java.cleanup.SimplifyBooleanExpressionVisitor`. I think it we could avoid requiring a dependency on `rewrite-static-analysis`.

src/test/resources/recipes/MultipleDereferencesRecipes.java
```diff
@@ -59,6 +59,7 @@ public J visitMethodInvocation(J.MethodInvocation elem, ExecutionContext ctx) {
                     if ((matcher = before.matcher(getCursor())).find()) {
                         doAfterVisit(new org.openrewrite.java.ShortenFullyQualifiedTypeReferences().getVisitor());
                         doAfterVisit(new org.openrewrite.java.cleanup.UnnecessaryParenthesesVisitor());
+                        doAfterVisit(new org.openrewrite.staticanalysis.SimplifyBooleanExpression().getVisitor());
```
knutwannheden, line 62: Rather than do a full visit, could we only visit the result or parent?

knutwannheden (approved):

src/main/java/org/openrewrite/java/template/RefasterTemplateProcessor.java
timtebeek: Thanks for pointing that out; seems we have _two_ full implementations, one in rewrite, and another in rewrite-static-analysis, likely because the second one can access a Kotlin marker that the first one cant. We should likely merge those two at some point.

src/test/resources/recipes/MultipleDereferencesRecipes.java
timtebeek: Not quite sure if that would work due to the recursive nature of the recipe, but we can log it as an improvement if you think it should.

PR #33, opened 2023-08-26
Make visitors extend `AbstractRefasterJavaVisitor`

knutwannheden:
This abstract base class declares some convenience methods so that the generated code gets easier to generate and read.


timtebeek (approved):

knutwannheden:
I will create a PR for the abstract base class. The idea with the suppliers is that the `JavaCompiler.Builder` doesn't get eagerly instantiated as soon as `getVisitor()` gets called. Instead this is delayed until it is actually used.

Issue #5, opened 2023-05-12
Detect and skip unsupported templates


knutwannheden:
Until the Refaster support is feature complete, it would still be nice to be able to use the annotation processor on a code base where some of the Refaster templates can be supported. For this to work the annotation processor must identify the templates which cannot be supported and then make sure to not generate anything for them (possibly even check if there is a corresponding generated file which should be deleted). It would also be nice to emit some messages to the developer, so it is clear why no recipe was generated.

knutwannheden:
@rpau The annotation processor now generates compiling code for most of Picnic's Refaster templates. I will investigate the remaining few issues and then we will need to decide how much of that we still want to address now and what we will leave as is.

timtebeek:
Perhaps good to use [the set of refaster rules developed at Picnic](https://github.com/PicnicSupermarket/error-prone-support/tree/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules) as a threshold of when we sufficiently exclude patterns we don't support yet. We could add the annotation processor there, if accepted, such that their `error-prone-contrib` module also becomes a rewrite recipe module.

Stephan202:
> We could add the annotation processor there, if accepted, such that their `error-prone-contrib` module also becomes a rewrite recipe module.

We're certainly open to this! (It's also on my/our TODO list, but speeding that along by submitting a proposal can't hurt :smile:.)

knutwannheden:
Currently the big limitations are:
- No support for generics
- No support for Refaster library functions or annotations other than BeforeTemplate and AfterTemplate

But we try to detect the cases we don't support and don't generate any code rather than wrong code, so I think we should already be able to get working recipes for a good subset of the rules.

timtebeek:
Played around with adding the annotation processor to error-prone-support just now; got initial recipe production going through:
`mvn clean compile -Dlicense.skipAddThirdParty=true -Dmaven.compiler.failOnError=false`
```diff
diff --git a/error-prone-contrib/pom.xml b/error-prone-contrib/pom.xml
index 75a31b47..67cda282 100644
--- a/error-prone-contrib/pom.xml
+++ b/error-prone-contrib/pom.xml
@@ -221,6 +221,12 @@
             <artifactId>testng</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.openrewrite</groupId>
+            <artifactId>rewrite-java</artifactId>
[...]

Issue #21, opened 2023-08-15
Support `return Refaster.anyOf(...)`


timtebeek:
## What problem are you trying to solve?
Right now we skip all refaster rules using any form of `Refaster.anyOf(...)`, whereas we could potentially cover `return Refaster.anyOf(...)` with less effort as compared to supporting `Refaster.anyOf(...)` anywhere in template for matches.
This will allow us to cover additional rules defined in [PicnicSupermarket/error-prone-support](https://github.com/PicnicSupermarket/error-prone-support), as discovered in https://github.com/openrewrite/rewrite-templating/issues/5#issuecomment-1678042625.

## Describe the solution you'd like
We already support multiple `@BeforeTemplate` annotated methods. We should expand support for multiple before statements to cover `return Refaster.anyOf` [as seen here](https://github.com/PicnicSupermarket/error-prone-support/blob/09208aa49a6d68ac7b75752b96f195d8c0ee12b5/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MapRules.java#L63-L67).
```java
@BeforeTemplate
boolean before(Map<K, V> map) {
  return Refaster.anyOf(map.keySet(), map.values(), map.entrySet()).isEmpty();
}
```

`Refaster.anyOf(...)` used anywhere else but in the return, repeatedly, or with preceding statements is left out of scope for now.

## Have you considered any alternatives or workarounds?
Not yet.

## Additional context
- https://github.com/openrewrite/rewrite-templating/issues/5
- https://github.com/PicnicSupermarket/error-prone-support
- https://errorprone.info/docs/refaster

timtebeek:
Had a brief look at adding this one over the weekend; the easiest case to tackle is likely something like this:
```java
public class RefasterAnyOf {
    @BeforeTemplate
    boolean before(String s) {
        return Refaster.anyOf(s.length() < 1, s.length() == 0);
    }

    @AfterTemplate
    boolean after(String s) {
        return s.isEmpty();
    }
}
```
And then ideally turn that into something similar to the multiple before statements we already know:
```
JavaTemplate before = Semantics.expression(this, "before", (String s) -> s.length() < 1).build();
JavaTemplate before2 = Semantics.expression(this, "before2", (String s) -> s.length() == 0).build();
JavaTemplate after = Semantics.expression(this, "after", (String s) -> s.isEmpty()).build();
```

[...]

PR #36, opened 2023-12-10
Print messages once, as to not flood console

timtebeek:
## What's changed?
- Only print notes once
- Assert there's no messages in tests
- Fix and enable disabled tests

## What's your motivation?
As per #5 I was looking to add this to error-prone-support, but noticed we flood the console with repeated messages there, mostly pending #21.

## Any additional context
- #5 
- #21 
- https://github.com/openrewrite/rewrite-migrate-java/commit/4bfadd20f38ec86ccbe80aaa77789c23d3ec3cd7

### DIFF
--- a/src/test/java/org/openrewrite/java/template/RefasterTemplateProcessorTest.java
+++ b/src/test/java/org/openrewrite/java/template/RefasterTemplateProcessorTest.java
@@ -47,6 +47,7 @@ class RefasterTemplateProcessorTest {
           .withClasspath(classpath())
           .compile(JavaFileObjects.forResource("refaster/" + recipeName + ".java"));
         assertThat(compilation).succeeded();
+        assertThat(compilation).hadNoteCount(0);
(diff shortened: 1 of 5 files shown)


timtebeek:
I think we can also close #5 ; we sufficiently skip unknown elements, which was the goal there.

joanvr (approved):

src/test/java/org/openrewrite/java/template/RefasterTemplateProcessorTest.java
```diff
@@ -55,17 +56,18 @@ void generateRecipe(String recipeName) {
 
     @ParameterizedTest
     @ValueSource(strings = {
-//      "ShouldSupportNestedClasses",
+      "ShouldSupportNestedClasses",
       "ShouldAddImports",
-//      "MultipleDereferences",
-//      "Matching",
```
knutwannheden, line 61: Do we need to disable these tests?

knutwannheden (approved):

PR #41, opened 2023-12-16
Skip Class generics

timtebeek:
## What's changed?
Add tests to show we ought to skip both Refaster (ok) and Class generics (not ok).

## What's your motivation?
OrOrElseThrows is found in error-prone-support, but breaks template generation there due to a compilation error.

There's also the issue that the test template only works on Java 11.
I wonder if we should use that as minimum for rewrite-templating.

Not quite sure yet how to catch and resolve this compilation error when using Java 11:
```
Compilation produced the following diagnostics:
Note: Generics are currently not supported
/home/tim/Documents/workspace/openrewrite/rewrite-templating/out/test/resources/refaster/OrOrElseThrow.java:26: error: incompatible types: bad return type in lambda expression
        return o1.orElseGet(() -> o2.orElseThrow());
                                                ^
    T cannot be converted to T
No files were generated.
```

### DIFF
--- a/src/test/java/com/google/errorprone/refaster/Refaster.java
+++ b/src/test/java/com/google/errorprone/refaster/Refaster.java
@@ -1,0 +1,22 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


timtebeek:
There's  something about the original recipe that still fails; but at least we can get these tests in already to ensure we do the right thing in these two cases, then the diff with the third case is easier to see.

knutwannheden (approved):

PR #43, opened 2023-12-16
Skip early when class generics detected

timtebeek:
## What's changed?
Add a test class that fails from error-prone-support.

## What's your motivation?
Resolve failure such that we can generate recipes there.


src/test/java/org/openrewrite/java/template/RefasterTemplateProcessorTest.java
```diff
@@ -57,6 +57,7 @@ void generateRecipe(String recipeName) {
     @ParameterizedTest
     @ValueSource(strings = {
       "OrElseGetGet",
+      "OrOrElseThrow",
```
timtebeek, line 60: When using Java 11 on my machine this produces
```
Compilation produced the following diagnostics:
Note: Generics are currently not supported
/home/tim/Documents/workspace/openrewrite/rewrite-templating/out/test/resources/refaster/OrOrElseThrow.java:26: error: incompatible types: bad return type in lambda expression
        return o1.orElseGet(() -> o2.orElseThrow());
                                                ^
    T cannot be converted to T
Note: Some messages have been simplified; recompile with -Xdiags:verbose to get full output
No files were generated.


	at org.openrewrite.java.template.RefasterTemplateProcessorTest.skipRecipeGeneration(RefasterTemplateProcessorTest.java:68)
```

timtebeek: Now confirmed to work in error-prone-support

timtebeek:
This fixes the build for 
- https://github.com/PicnicSupermarket/error-prone-support/compare/master...timtebeek:error-prone-support:add-rewite-templating
- https://github.com/openrewrite/rewrite-templating/issues/5#issuecomment-1858871230

PR #45, opened 2023-12-16
Reduce warnings by generating JavaDoc

timtebeek:
## What's changed?
Also generate a line of JavaDoc for classes.

## What's your motivation?
Reduce warnings on use in error-prone-support.

## Any additional context
- #5

### DIFF
--- a/src/main/java/org/openrewrite/java/template/processor/TemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/processor/TemplateProcessor.java
@@ -215,8 +215,8 @@ public class TemplateProcessor extends TypeAwareProcessor {
                                 }
 
                                 out.write("\n");
-                                out.write("public class " + templateFqn.substring(templateFqn.lastIndexOf('.') + 1) + " {\n");
-                                out.write("    public static JavaTemplate.Builder getTemplate() {\n");
+                                out.write("class " + templateFqn.substring(templateFqn.lastIndexOf('.') + 1) + " {\n");
+                                out.write("    static JavaTemplate.Builder getTemplate() {\n");
                                 out.write("        return JavaTemplate\n");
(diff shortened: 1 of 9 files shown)


src/main/java/org/openrewrite/java/template/processor/TemplateProcessor.java
```diff
@@ -215,8 +215,8 @@ public void visitIdent(JCTree.JCIdent ident) {
                                 }
 
                                 out.write("\n");
-                                out.write("public class " + templateFqn.substring(templateFqn.lastIndexOf('.') + 1) + " {\n");
-                                out.write("    public static JavaTemplate.Builder getTemplate() {\n");
+                                out.write("class " + templateFqn.substring(templateFqn.lastIndexOf('.') + 1) + " {\n");
+                                out.write("    static JavaTemplate.Builder getTemplate() {\n");
```
timtebeek, line 219: This class and method are accessed from the PatternBuilder. I don't expect issues with the reduced visibility, but figured ask for a second opinion just in case.
https://github.com/openrewrite/rewrite-templating/blob/022373a047323b6a401d13adc986284f290363ea/src/main/java/org/openrewrite/java/template/internal/PatternBuilder.java#L30-L39

timtebeek: Alternatively we had plans in the past to merge this into `RefasterTemplateProcessor`, which might make sense now too as I don't think we've come to use these templates separately.

timtebeek: Just confirmed to work against `apache/maven-build-cache-extension` with a local build; figured that makes this a low impact enough change that we can see this through and at worst refine on the main branch later.

PR #50, opened 2023-12-17
Generate default constructor with JavaDoc for Xdoclint

timtebeek:

### DIFF
--- a/src/main/java/org/openrewrite/java/template/processor/TemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/processor/TemplateProcessor.java
@@ -215,7 +215,7 @@ public class TemplateProcessor extends TypeAwareProcessor {
                                 }
 
                                 out.write("\n");
-                                out.write("/**\n * OpenRewrite `" + templateName.getValue() + "` template created for  `" + templateFqn.split("_")[0] + "`.\n */\n");
+                                out.write("/**\n * OpenRewrite `" + templateName.getValue() + "` template created for {@code " + templateFqn.split("_")[0] + "}.\n */\n");
                                 String templateClassName = templateFqn.substring(templateFqn.lastIndexOf('.') + 1);
                                 out.write("public class " + templateClassName + " {\n");
                                 out.write("    /**\n");
@@ -223,8 +223,8 @@ public class TemplateProcessor extends TypeAwareProcessor {
                                 out.write("     */\n");
                                 out.write("    public " + templateClassName + "() {}\n\n");
                                 out.write("    /**\n");
-                                out.write("     * Get the `JavaTemplate.Builder` to match or replace.\n");
-                                out.write("     * @return the `JavaTemplate.Builder`.\n");
+                                out.write("     * Get the {@code JavaTemplate.Builder} to match or replace.\n");
+                                out.write("     * @return the JavaTemplate builder.\n");
                                 out.write("     */\n");
                                 out.write("    public static JavaTemplate.Builder getTemplate() {\n");
                                 out.write("        return JavaTemplate\n");
--- a/src/main/java/org/openrewrite/java/template/processor/TemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/processor/TemplateProcessor.java
@@ -216,8 +216,16 @@ public class TemplateProcessor extends TypeAwareProcessor {
 
                                 out.write("\n");
                                 out.write("/**\n * OpenRewrite `" + templateName.getValue() + "` template created for  `" + templateFqn.split("_")[0] + "`.\n */\n");
-                                out.write("public class " + templateFqn.substring(templateFqn.lastIndexOf('.') + 1) + " {\n");
-                                out.write("    /**\n     * @return `JavaTemplate` to match or replace.\n     */\n");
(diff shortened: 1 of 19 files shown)


src/test/resources/template/ParameterReuseRecipe$1_before.java
```diff
@@ -16,7 +16,19 @@
 package foo;
 import org.openrewrite.java.*;
 
+/**
+ * OpenRewrite `before` template created for  `foo.ParameterReuseRecipe$1`.
+ */
 public class ParameterReuseRecipe$1_before {
+    /**
+     * Instantiates a new {@link ParameterReuseRecipe$1_before} instance.
+     * @return the new instance
```
Stephan202, line 25: `@return` doesn't apply (or at least, isn't necessary) for constructors.

src/test/resources/template/ParameterReuseRecipe$1_before.java
```diff
@@ -16,7 +16,19 @@
...
+     */
+    public ParameterReuseRecipe$1_before() {}
+
+    /**
+     * @return `JavaTemplate` to match or replace.
```
Stephan202, line 30: It _might_ be that doclint will complain here about the absence of a summary line, but not sure.

src/test/resources/template/ParameterReuseRecipe$1_before.java
```diff
@@ -16,7 +16,19 @@
 package foo;
 import org.openrewrite.java.*;
 
+/**
+ * OpenRewrite `before` template created for  `foo.ParameterReuseRecipe$1`.
+ */
 public class ParameterReuseRecipe$1_before {
+    /**
+     * Instantiates a new {@link ParameterReuseRecipe$1_before} instance.
```
Stephan202, line 24: There's also a doclint check that validates `@link`s; I'm not sure, but it could be that the `$` should be a `.` or `#`.

Stephan202 (commented):
Nice! Left two comments about possible further "roadblocks", but if you already tested this against the other repo: LGTM!

src/test/resources/template/ParameterReuseRecipe$1_before.java
timtebeek: Thanks! Removed.

src/test/resources/template/ParameterReuseRecipe$1_before.java
timtebeek: Indeed it did; fixed with an extra line.

src/test/resources/template/ParameterReuseRecipe$1_before.java
timtebeek: Good to know! Removed the `@link`; life's too short.

timtebeek:
We need the same documented default constructor still for RefasterTemplateProcessor generated Recipe & Recipes.

src/main/java/org/openrewrite/java/template/processor/TemplateProcessor.java
```diff
@@ -216,8 +216,16 @@ public void visitIdent(JCTree.JCIdent ident) {
 
                                 out.write("\n");
                                 out.write("/**\n * OpenRewrite `" + templateName.getValue() + "` template created for  `" + templateFqn.split("_")[0] + "`.\n */\n");
```
Stephan202, line 218: Double space:
```suggestion
                                out.write("/**\n * OpenRewrite `" + templateName.getValue() + "` template created for `" + templateFqn.split("_")[0] + "`.\n */\n");
```

src/test/resources/template/ParameterReuseRecipe$1_before.java
```diff
@@ -16,7 +16,19 @@
...
+     * @return the new instance
+     */
```
Stephan202, line 26: Looks like these files were not updated. (But not sure they need to be?)

Stephan202 (approved):

src/test/resources/template/ParameterReuseRecipe$1_before.java
```diff
@@ -16,7 +16,19 @@
 package foo;
 import org.openrewrite.java.*;
 
+/**
+ * OpenRewrite `before` template created for  `foo.ParameterReuseRecipe$1`.
```
Stephan202, line 20: NB: Since this is JavaDoc it'd be more accurate to use `{@code ...}` rather than backticks. (Same applies to #51.)

Stephan202 (commented):
One more thought; up to you :)

src/test/resources/template/ParameterReuseRecipe$1_before.java
timtebeek: Yeah funnily enough `com.google.testing.compile:compile-testing` ignores any whitespace or comments when comparing if the outcome is the same, so I've removed the comments from the output again is it's hard to keep in sync. Thanks for pointing out the differences though! That's what prompted me to just remove those here.

src/test/resources/template/ParameterReuseRecipe$1_before.java
timtebeek: Thanks! Applied here already. Now on to #51.

PR #51, opened 2023-12-17
Extract display name and description from JavaDoc

timtebeek:

### DIFF
--- a/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
@@ -405,7 +405,7 @@ public class RefasterTemplateProcessor extends TypeAwareProcessor {
                 Tokens.Comment comment = cu.docComments.getComment(classDecl);
                 if (comment != null && comment.getText() != null && !comment.getText().isEmpty()) {
                     String commentText = comment.getText()
-                            .replaceAll("\\{@\\S+\\s+(.*)}", "`$1`")
+                            .replaceAll("\\{@\\S+\\s+(.*?)}", "`$1`")
                             .replace("\\", "\\\\")
                             .replace("\"", "\\\"")
                             .replace("\b", "\\b")
--- a/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
@@ -414,8 +414,14 @@ public class RefasterTemplateProcessor extends TypeAwareProcessor {
                             .replace("\r", "\\r");
                     String[] lines = commentText.split("\\R+", 2);
                     displayName = lines[0].trim();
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
```diff
@@ -399,6 +400,27 @@ private String recipeDescriptor(JCTree.JCClassDecl classDecl, String defaultDisp
                 String displayName = defaultDisplayName;
                 String description = defaultDescription;
                 Set<String> tags = new LinkedHashSet<>();
+
+                // Extract from JavaDoc
+                Tokens.Comment comment = cu.docComments.getComment(classDecl);
+                if (comment != null && comment.getText() != null && !comment.getText().isEmpty()) {
+                    String commentText = comment.getText()
+                            .replace("{@link ", "`")
+                            .replace("}", "`")
```
Stephan202, line 409: I guess `{@code` should get the same treatment.

(This wouldn't work for `{@value`, but it seems highly unlikely one would use that for a Refaster rule.)

Stephan202 (commented):
Nice! Within Error Prone Support we currently have [similar functionality](https://github.com/PicnicSupermarket/error-prone-support/blob/7529b9925145c3fdaf0864a6df99b6cbb20d7d7f/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java#L390-L391) using an `@Description` annotation, but I have pondered in the past whether indeed we should instead just look at the Javadoc. So this change SGTM!

timtebeek:
> Nice! Within Error Prone Support we currently have [similar functionality](https://github.com/PicnicSupermarket/error-prone-support/blob/7529b9925145c3fdaf0864a6df99b6cbb20d7d7f/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java#L390-L391) using an `@Description` annotation, but I have pondered in the past whether indeed we should instead just look at the Javadoc. So this change SGTM!

Ah interesting! Had been looking for something like that on your end, but didn't immediately spot anything. I only see that `@tech.picnic.errorprone.refaster.annotation.Description` [applied twice](https://github.com/search?q=repo%3APicnicSupermarket%2Ferror-prone-support+%22%40Description%22&type=code) though, once in tests. I'm guessing you're considering just taking the JavaDoc there as well then?

src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
timtebeek: Thanks! Picked up in https://github.com/openrewrite/rewrite-templating/pull/51/commits/fc3bb8d87c57c47fd35306d36588d1b231880ddf to replace any such tags.

Stephan202:
> I'm guessing you're considering just taking the JavaDoc there as well then?

Nothing decided, but at this point it seems likely: the JavaDoc seems more maintainable and easy to use, especially when considering line wrapping.

PR #52, opened 2023-12-18
Add `@SuppressWarnings("all")` to stop scanning tool complaints

timtebeek:

### DIFF
--- a/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
@@ -154,7 +154,10 @@ public class RefasterTemplateProcessor extends TypeAwareProcessor {
                     }
 
                     for (Set<String> imports : imports.values()) {
-                        imports.removeIf(i -> "java.lang".equals(i.substring(0, i.lastIndexOf('.'))));
+                        imports.removeIf(i -> {
+                            int endIndex = i.lastIndexOf('.');
+                            return endIndex < 0 || "java.lang".equals(i.substring(0, endIndex));
+                        });
                         imports.remove(BEFORE_TEMPLATE);
                         imports.remove(AFTER_TEMPLATE);
                     }
@@ -179,6 +182,7 @@ public class RefasterTemplateProcessor extends TypeAwareProcessor {
                     String refasterRuleClassName = classDecl.sym.fullname.toString().substring(classDecl.sym.packge().fullname.length() + 1);
                     recipe.append("/**\n * OpenRewrite recipe created for Refaster template {@code ").append(refasterRuleClassName).append("}.\n */\n");
                     String recipeName = templateFqn.substring(templateFqn.lastIndexOf('.') + 1);
+                    recipe.append("@javax.annotation.Generated(\"org.openrewrite.java.template.processor.RefasterTemplateProcessor\")\n");
                     recipe.append("@NonNullApi\n");
                     recipe.append(descriptor.classDecl.sym.outermostClass() == descriptor.classDecl.sym ?
                             "public class " : "public static class ").append(recipeName).append(" extends Recipe {\n\n");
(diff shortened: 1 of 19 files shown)


src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
```diff
@@ -340,6 +344,7 @@ public void visitClassDef(JCTree.JCClassDecl classDecl) {
                             if (outerClassRequired) {
                                 out.write("/**\n * OpenRewrite recipes created for Refaster template {@code " + inputOuterFQN + "}.\n */\n");
                                 String outerClassName = className.substring(className.lastIndexOf('.') + 1);
+                                out.write("@javax.annotation.Generated(\"org.openrewrite.java.template.processor.RefasterTemplateProcessor\")\n");
```
timtebeek, line 347: When added here it's technically not necessary to _also_ add it to each nested class, but that then requires switching around a bunch of code for a harmless annotation. Figured point it out but don't feel like taking that on, hope you all agree.

timtebeek: In a similar vein I didn't feel like handling imports, so it's a one liner here.

timtebeek:
@stephan202 When running this branch against error-prone-support I do see a rather curious error
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.10.1:testCompile (default-testCompile) on project refaster-runner: Compilation failure: Compilation failure: 
[ERROR] /home/tim/Documents/workspace/timtebeek/error-prone-support/refaster-runner/target/generated-test-sources/test-annotations/tech/picnic/errorprone/refaster/runner/FooRulesRecipes.java:[25,18] cannot find symbol
[ERROR]   symbol:   class Generated
[ERROR]   location: package javax.annotation
```
1. somehow the `RefasterTemplateProcessor` configured in `error-prone-contrib` is active on `refaster-runner`
2. the `javax.annotation.Generated` annotation can not be found there; should we use a different one?

Any ideas there? Wouldn't want to release a version that breaks things on error-prone-support, or requires undesired changes there.

Stephan202:
> somehow the `RefasterTemplateProcessor` configured in `error-prone-contrib` is active on `refaster-runner`

Yeah, that's because I [moved the annotation processor declaration](https://github.com/PicnicSupermarket/error-prone-support/pull/925#discussion_r1429522961) :)

> the `javax.annotation.Generated` annotation can not be found there; should we use a different one?

Indeed, `javax.annotation.Generated` is no longer present on newer JDKs, where it's replaced with `javax.annotation.processing.Generated` (since JDK 9). IIRC this had something to do with splitting up the `javax.annotation` package to comply with the module system. Since you also target JDK 8, this can get a bit messy. Two other options:
1. Define a custom `@Generated` annotation. Tools such as Error Prone [support](https://github.com/google/error-prone/blob/d272dfa8da5f6ed09bb97acc153f9a7169185ab7/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java#L1663) this, and there's precedent set by e.g. [Lombok](https://github.com/projectlombok/lombok/blob/00893744eb56d0e44fa7f59ea28dd3e0da8ff0e8/src/core/lombok/Generated.java).
2. Instead add `@SuppressWarnings("all")`. That matches what [Lombok](https://github.com/search?q=repo%3Aprojectlombok%2Flombok%20suppresswarnings%20all&type=code) and [Immutables](https://github.com/search?q=repo%3Aimmutables%2Fimmutables+suppresswarnings+all&type=code) do, and since google/error-prone@47f0c72513801eeb2ef19b07f2e3dbdc8414668b this is also supported by Error Prone.

timtebeek:
Helpful context once again, thanks! Switched to `@SuppressWarnings("all")` as that seemed easiest: no additional API surface or dependency. Thanks!

timtebeek:
Does that also mean I can drop the empty constructors and Javadoc again? 😇

Stephan202 (approved):
Great question! I _suspect_ that Doclint doesn't respect `@SuppressWarnings("all")`, but you could try :smile:

Issue #47, opened 2023-12-17
Support additional Refaster features seen in error-prone-support


timtebeek:
## What problem are you trying to solve?
Expand the set of Refaster rules in Error Prone Support that we cover with OpenRewrite recipes.

## Describe the solution you'd like
Support the following cases not currently covered.
- [ ] Generics as seen in [ImmutableSortedMapRules.java](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ImmutableSortedMapRules.java)
- [ ] `com.google.errorprone.refaster.Refaster` as seen in [EqualityRules.java](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/EqualityRules.java)
  - #21 
- [ ] `@UseImportPolicy` as seen in [MockitoRules.java](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MockitoRules.java)
- [ ] `@Placeholder` as seen in [LongStreamRules.java](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/LongStreamRules.java)
- [ ] `@AlsoNegation` as seen in [BigDecimalRules.java](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java)
- [ ] `@DoNotCall` as seen in [TestNGToAssertJRules.java](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TestNGToAssertJRules.java)
- [ ] `@Repeated` as seen in [WebClientRules.java](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/WebClientRules.java)
- [ ] `@Matches` as seen in [ReactorRules.java](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java)

Stephan202:
[...]

PR #56, opened 2023-12-25
Support Arrays and ignore if statements

timtebeek:
Pulled out of #53 for easier review and to minimize diff.

### DIFF
--- a/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
@@ -494,15 +494,9 @@ public class RefasterTemplateProcessor extends TypeAwareProcessor {
 
             private void maybeRemoveImports(Map<JCTree.JCMethodDecl, Set<String>> importsByTemplate, StringBuilder recipe, JCTree.JCMethodDecl beforeTemplate, JCTree.JCMethodDecl afterTemplate) {
                 Set<String> beforeImports = getBeforeImportsAsStrings(importsByTemplate, beforeTemplate);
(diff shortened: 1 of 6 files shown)


src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
```diff
@@ -811,17 +806,16 @@ private boolean resolve(Context context, JCCompilationUnit cu) {
...
-    private final Set<String> printedMessages = new HashSet<>();
+    private final Map<String, Integer> printedMessages = new TreeMap<>();
 
     /**
      * @param message The message to print
      * @param symbol  The symbol to attach the message to; printed as clickable link to file
      */
     private void printNoteOnce(String message, Symbol.ClassSymbol symbol) {
-        if (printedMessages.add(message)) {
+        if (printedMessages.compute(message, (k, v) -> v == null ? 1 : v + 1) == 1) {
```
pstreef, line 818: why this change? Is this somehow more performant?

pstreef (approved):

src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
timtebeek: It's to allow one to inspect the number of times recipes are excluded, as used in this https://github.com/openrewrite/rewrite-templating/issues/47#issuecomment-1868511830

Didn't want to make that part of the standard output though; I think once and linked to the source is enough, as opposed to just a summary of how often there are exclusions not tied to a particular instance. Having this counter allows us to easily print it as a one off though.

PR #57, opened 2023-12-27
Inline Javatemplate in RefasterTemplateProcessor

timtebeek:
WIP to replace directly create JavaTemplate in RefasterTemplateProcessor instead of indirection.


timtebeek:
@knutwannheden As expected `TemplateProcessorTest` now fails, as we no longer generate the `Semantics` statements that operates on; how would you like to handle that?

- Should we adapt the test inputs to use `Semantics` directly to trigger similar output?
- Should we adapt `TemplateProcessor` to read `@Before/AfterTemplate` annotated methods?
- Should we retain `PatternBuilder` at all now that the indirection isn't techically needed anymore?

src/test/resources/refaster/ArraysRecipe.java
```diff
@@ -53,8 +60,12 @@ public String getDescription() {
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
         JavaVisitor<ExecutionContext> javaVisitor = new AbstractRefasterJavaVisitor() {
-            final JavaTemplate before = Semantics.expression(this, "before", (String[] strings) -> String.join(", ", strings)).build();
-            final JavaTemplate after = Semantics.expression(this, "after", (String[] strings) -> String.join(":", strings)).build();
+            final JavaTemplate before = JavaTemplate
+                    .builder("String.join(\", \", foo.Arrays.before.strings)")
```
knutwannheden, line 64: This `foo.Arrays.before.strings` looks weird. Is that a parameter?

timtebeek: Indeed the parameters look weird there for the test I added recently
https://github.com/openrewrite/rewrite-templating/blob/7993e10ca5ba483d4b56ac21d89a9811a3cfef9b/src/test/resources/refaster/Arrays.java#L21-L31

src/test/resources/refaster/EscapesRecipes.java
```diff
@@ -76,8 +91,14 @@ public String getDescription() {
         @Override
         public TreeVisitor<?, ExecutionContext> getVisitor() {
             JavaVisitor<ExecutionContext> javaVisitor = new AbstractRefasterJavaVisitor() {
-                final JavaTemplate before = Semantics.expression(this, "before", (String value) -> String.format("\"%s\"", com.sun.tools.javac.util.Convert.quote(value))).build();
-                final JavaTemplate after = Semantics.expression(this, "after", (String value) -> com.sun.tools.javac.util.Constants.format(value)).build();
+                final JavaTemplate before = JavaTemplate
+                        .builder("String.format(\"\\\"%s\\\"\", com.sun.tools.javac.util.Convert.quote(foo.Escapes.ConstantsFormat.before.value))")
```
knutwannheden, line 95: The `foo.Escapes.ConstantsFormat.before.value` parameter here also looks wrong. Possibly something like this is required? https://github.com/openrewrite/rewrite-templating/blob/b138ac0fa33ccde1fc5305a484bc68b4ab9782d0/src/main/java/org/openrewrite/java/template/processor/TemplateProcessor.java#L97-L101

src/test/resources/template/ShouldAddClasspathRecipes.java
```diff
@@ -0,0 +1,58 @@
...
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.JavaTemplate;
+import org.openrewrite.java.template.Semantics;
+
+import static java.util.regex.Pattern.DOTALL;
+import static org.slf4j.LoggerFactory.getLogger;
+
+public class ShouldAddClasspathRecipes {
```
timtebeek, line 28: Probably best to pull this into a separate PR such that we can test and separate the two processors better. Would not have to run both processors to generate the output then for these tests.

src/test/resources/refaster/EscapesRecipes.java
timtebeek: Thanks! Finally managed to get it working, which unblocks running recipes that use Java 11+ constructs on Java 8. 🎉

knutwannheden (approved):
Nice that we finally got this working!

src/test/resources/refaster/ArraysRecipe.java
```diff
@@ -53,8 +60,12 @@ public String getDescription() {
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
         JavaVisitor<ExecutionContext> javaVisitor = new AbstractRefasterJavaVisitor() {
-            final JavaTemplate before = Semantics.expression(this, "before", (String[] strings) -> String.join(", ", strings)).build();
-            final JavaTemplate after = Semantics.expression(this, "after", (String[] strings) -> String.join(":", strings)).build();
+            final JavaTemplate before = JavaTemplate
+                    .builder("String.join(\", \", #{strings:any(java.lang.String[])})")
```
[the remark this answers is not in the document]
knutwannheden: I see, we have an explicit exception for `java.lang`: https://github.com/openrewrite/rewrite-templating/blob/261ad17a7393fa5db5553b8a58e3a646343ee4ea/src/main/java/org/openrewrite/java/template/internal/TemplateCode.java#L118

It does make the template code a bit more readable, but it could cause problems when the Java class an after template gets embedded into is in a package which has a class by the same name, which is unusual but not unheard of.

I am a bit on the fence on this one. `java.lang` is a special case also in the `ShortenFullyQualifiedTypeReferences` recipe (for basically the same reason). But I think in probably around 90% of the cases the name would end up getting shortened, because the compilation unit already has another (unqualified) reference to the same type in which case the recipe knows it can shorten the reference.

As I pointed out before, this isn't directly related to your PR. It is just something that caught my eye.

src/test/resources/refaster/ArraysRecipe.java
```diff
@@ -53,8 +60,12 @@ public String getDescription() {
...
+                    .build();
+            final JavaTemplate after = JavaTemplate
+                    .builder("String.join(\":\", #{strings:any(java.lang.String[])})")
+                    .build();
```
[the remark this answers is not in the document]
timtebeek: Yes the main motivator for me was PRs like this one that aim to use Java 9+ constructs in recipes, which previously meant that code was in the generated recipes as well(likely?) bumping the Java version required to run the recipe.
- https://github.com/openrewrite/rewrite-migrate-java/pull/399

With this change I think we should be good to have the generated recipes included with the other Java 8 compatible classes.

src/test/resources/refaster/ArraysRecipe.java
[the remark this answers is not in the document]
timtebeek: Yes we can tackle this separately; there's a few other items too I'd been looking to do after this lands, since any changes previously were likely to conflict with this one.

PR #59, opened 2024-01-02
Remove parentheses when after is a unary too

timtebeek:
As discovered on
- https://github.com/openrewrite/rewrite-static-analysis/pull/231


src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
```diff
@@ -257,7 +257,7 @@ public void visitClassDef(JCTree.JCClassDecl classDecl) {
                             maybeRemoveImports(staticImports, recipe, entry.getValue(), descriptor.afterTemplate);
 
                             List<String> embedOptions = new ArrayList<>();
-                            if (getType(descriptor.afterTemplate) == JCTree.JCParens.class) {
+                            if (getType(descriptor.afterTemplate) == JCTree.JCParens.class || getType(descriptor.afterTemplate) == JCTree.JCUnary.class) {
```
timtebeek, line 260: Technically this matches slightly too wide; in practice I don't expect any of the other unary recipes to be used in `after`, or running `UnnecessaryParenthesesVisitor` to cause any issues when they do occur.

PR #60, opened 2024-01-03
Decouple TemplateProcessorTest from RefasterTemplateProcessor

timtebeek:
Ahead of 
- #57

I'd considered also changing the tests to no longer use files as in and output but text blocks, but figured we can discuss and pick that up separately.


src/test/resources/template/ParameterReuseRecipe.java
```diff
@@ -15,20 +15,14 @@
...
-    }
-
-    @AfterTemplate
-    boolean after() {
-        return true;
-    }
+public class ParameterReuseRecipe {
+    JavaIsoVisitor visitor = new JavaIsoVisitor<ExecutionContext>() {
+        JavaTemplate.Builder before = Semantics.expression(this, "before", (String s) -> s.equals(s));
+        JavaTemplate.Builder after = Semantics.expression(this, "after", (String s) -> true);
```
timtebeek, line 26: The goal here is to more closely match the use in rewrite-micrometer, such that we can troubleshoot that more easily:
https://github.com/openrewrite/rewrite-micrometer/blob/4e4e71a0b2e8dd78477a4fcac711d6048dc3577c/src/main/java/org/openrewrite/micrometer/misk/NoExplicitEmptyLabelList.java#L50-L57

timtebeek:
Going to merge this one as I've only changes classes in src/test, such that I can build on this for a hopeful fix for the missing TemplateProcessor classpath entry.

PR #61, opened 2024-01-06
Missing template classpath entry for parameters

timtebeek:
Turns out we were indeed missing classpath entries, but only for the parameters, as seen in:
https://github.com/openrewrite/rewrite-micrometer/actions/workflows/ci.yml

TemplatCode.process gets the body passed in separately from the parameters; and we only looked at the body for classpath entries and imports. Now we also add classpath entries for input parameters (but no imports).


timtebeek:
After a local install it seems to work in
1. rewrite-micrometer
2. rewrite-migrate-java

timtebeek:
Also works in error-prone-support if combined with 
- https://github.com/openrewrite/rewrite-templating/pull/62

knutwannheden (approved):

PR #62, opened 2024-01-06
Fix compilation error around throws on EPS

timtebeek:
Before this change we produced the following, which does not compile due to the new lines in the String.
```java
public static JavaTemplate.Builder getTemplate() {
    return JavaTemplate
            .builder("{
            throw new AssertionError();
        }");
}
```

### DIFF
--- a/src/main/java/org/openrewrite/java/template/internal/TemplateCode.java
+++ b/src/main/java/org/openrewrite/java/template/internal/TemplateCode.java
@@ -37,7 +37,10 @@ public class TemplateCode {
             StringBuilder builder = new StringBuilder("JavaTemplate\n");
             builder
                     .append("    .builder(\"")
-                    .append(writer.toString().replace("\\", "\\\\").replace("\"", "\\\""))
+                    .append(writer.toString()
+                            .replace("\\", "\\\\")
+                            .replace("\"", "\\\"")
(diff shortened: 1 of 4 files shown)


src/main/java/org/openrewrite/java/template/internal/TemplateCode.java
```diff
@@ -37,7 +37,10 @@ public static <T extends JCTree> String process(T tree, List<JCTree.JCVariableDe
             StringBuilder builder = new StringBuilder("JavaTemplate\n");
             builder
                     .append("    .builder(\"")
-                    .append(writer.toString().replace("\\", "\\\\").replace("\"", "\\\""))
+                    .append(writer.toString()
+                            .replace("\\", "\\\\")
+                            .replace("\"", "\\\"")
+                            .replaceAll("\\R", " "))
```
knutwannheden, line 43: I would rather replace them with an escaped line terminator (`\n`) because in a text block they are significant whitespace.

timtebeek: Better indeed; done in https://github.com/openrewrite/rewrite-templating/pull/62/commits/367c450e1eb19d77182b3fde965763f382b93b92

PR #63, opened 2024-01-09
Improve performance for `J.Ternary` expressions

knutwannheden:


knutwannheden:
I tested this against the rewrite-static-analysis recipe and it still isn't quite right. Trying to figure it out now.

knutwannheden:
For performance reasons we might also want to generate more other specific `visit()` methods like `visitTernary()` here.

timtebeek (approved):

PR #68, opened 2024-01-27
Don't skip embedOptions for parameterless recipes

### LINKED ISSUE #66: missing SHORTEN_NAMES for parameterless recipes

```java
        @BeforeTemplate
        String before() {
            return "foo";
        }

        @AfterTemplate
        String after() {
            return SomeClass.SOME_CONSTANT;
        }
```

## What did you expect to see?
The generated recipe should contain SHORTEN_NAMES. It's meant to be generated with _every_ recipe, according to https://github.com/openrewrite/rewrite-templating/blob/0e4a1440df56537041bbc86294dc77f22758c288/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java#L272

## What did you see instead?

No additional recipe references. Therefore running the recipe leads to a FQN for `SomeClass.SOME_CONSTANT`. (Workaround: Add the `ShortenFullyQualifiedTypeReferences` to the recipe list explicitly).

Root cause is https://github.com/openrewrite/rewrite-templating/blob/0e4a1440df56537041bbc86294dc77f22758c288/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java#L278 not adding the `embedOptions` collection (it does this correctly in https://github.com/openrewrite/rewrite-templating/blob/main/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java#L284).

Yes. I would combine the 2 code blocks to remove the code duplication, since such code duplications will lead to similar failures again.

Bananeweizen:
Deduplicate the logic for creating the recipe code to avoid an inconcistency in how the embedOptions are used.

Fixes #66.

### Checklist
- [X] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [X] I've used the IntelliJ IDEA auto-formatter on affected files

### DIFF
--- a/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
@@ -274,16 +274,16 @@ public class RefasterTemplateProcessor extends TypeAwareProcessor {
                                 embedOptions.add("SIMPLIFY_BOOLEANS");
                             }
 
-                            if (parameters.isEmpty()) {
-                                recipe.append("                    return embed(").append(after).append(".apply(getCursor(), elem.getCoordinates().replace()), getCursor(), ctx);\n");
-                            } else {
-                                recipe.append("                    return embed(\n");
-                                recipe.append("                            ").append(after).append(".apply(getCursor(), elem.getCoordinates().replace(), ").append(parameters).append("),\n");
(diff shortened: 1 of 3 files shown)


knutwannheden (approved):
LGTM. Thanks for the fix!

PR #70, opened 2024-01-28
enforce UTF-8 resources on older JVMs

### LINKED ISSUE #69: Test fails only in IntelliJ

Open `RefasterTemplateProcessorTest.java`, hit the Run button of IntelliJ. On my Windows 10 machine this leads to a failing test.

## What did you see instead?

```
> Difference in expected tree and actual tree.
      Expected node: Line 56 COMPILATION_UNIT->CLASS(UseStringIsEmptyRecipe)->METHOD(getDescription)->BLOCK(non-static)->RETURN->STRING_LITERAL(Second line that should show up in description only.
     May contain " and ' and \" and \\" and \n.
     Or even references to `String`.
     Or unicode ðŸ�›.)
      Actual node: Line 41 COMPILATION_UNIT->CLASS(UseStringIsEmptyRecipe)->METHOD(getDescription)->BLOCK(non-static)->RETURN->STRING_LITERAL(Second line that should show up in description only.
     May contain " and ' and \" and \\" and \n.
     Or even references to `String`.
     Or unicode ðŸ?›.)
```

The mismatch is in the unicode "bug" character: 🐛 from https://github.com/openrewrite/rewrite-templating/blob/0e4a1440df56537041bbc86294dc77f22758c288/src/test/resources/refaster/UseStringIsEmpty.java#L28. Please note that it is not correctly decoded in both the expected and actual values.
Running the same test (in the same git repository) via gradlew command line is fine.

I use IntelliJ too seldom to really know what I could investigate here, so probably not.

Bananeweizen:
Before Java 18 the default charset was platform dependent, which leads to failing tests on older JVMs with another default charset.

## What's changed?
For unit tests all resources are now read as UTF-8 via the system property file.encoding.

## What's your motivation?
Fixes #69.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -97,6 +97,8 @@ dependencies {
 
 tasks.withType<Test> {
     useJUnitPlatform()
+    // enforce reading resources as UTF-8 also on JDKs before Java 18
+    systemProperty("file.encoding", "UTF-8")
 }
 
 tasks.withType<Javadoc> {


timtebeek (approved):
Good to see this fixed, thanks!

PR #71, opened 2024-02-02
Fix multiline values in `@RecipeDescriptor`

JLLeitschuh:
Signed-off-by: Jonathan Leitschuh <Jonathan.Leitschuh@gmail.com>

## What's changed?

Fixed the generated code from not properly escaping all special characters for Java.

This was done by using https://commons.apache.org/proper/commons-text/javadocs/api-release/org/apache/commons/text/StringEscapeUtils.html#escapeJava(java.lang.String)

## What's your motivation?

Discovered this bug in https://github.com/openrewrite/rewrite-static-analysis/pull/252
This fixes it.

I know that Jonathan's opinion is to try to minimize the number of third party libraries that openrewrite depends upon, but
given that the purpose of this codebase is to do text generation/manipulation, this library seems appropreate for the use-case.

Also, if this templating engine is ever used to generate code for other languages, the ability to escape those languages will likely also be appropreate.

I could copy-paste the escaping logic from `commons-text` into this codebase, but I figured using an industry standard library for this sort of thing
would better stand the test of time.

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ build/
 .boot-releases
 out/
 .attach_pid*
+.DS_Store
--- a/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
@@ -15,7 +15,10 @@
  */
 package org.openrewrite.java.template.processor;
 
-import com.sun.source.tree.*;
+import com.sun.source.tree.AnnotationTree;
+import com.sun.source.tree.MethodTree;
+import com.sun.source.tree.Tree;
+import com.sun.source.tree.VariableTree;
 import com.sun.tools.javac.code.Symbol;
 import com.sun.tools.javac.code.Type;
 import com.sun.tools.javac.code.TypeTag;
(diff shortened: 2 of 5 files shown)


build.gradle.kts
```diff
@@ -76,6 +76,7 @@ val tools = compiler.get().metadata.installationPath.file("lib/tools.jar")
 dependencies {
     compileOnly(files(tools))
     compileOnly("org.jetbrains:annotations:24.0.+")
+    implementation("org.apache.commons:commons-text:latest.release")
```
timtebeek, line 79: Hmm; as this is an annotation processor we had intentionally kept dependencies to an absolute minimum; while I understand the appeal of a library; would it be possible to sanitize without this lib?

src/test/resources/refaster/MultilineAnnotation.java
JLLeitschuh, line 1: I think it would be a pretty fun exercise to write an OpenRewrite recipe that attempted to find all test data/test strings that used cultural references 😆

timtebeek (commented):
Thanks for figuring out what plagued your other PR; small request below, but otherwise don't see any hurdles.

build.gradle.kts
JLLeitschuh: Probably, but I'd rather just vendor the dependency with the shadowjar plugin

JLLeitschuh: Resolved

timtebeek (approved):
Thanks a lot!

PR #77, opened 2024-03-08
Generate more precise `UsesMethod` predicates

knutwannheden:
In the case of overloads it could be interesting to generate more precise `UsesMethod` predicates, as this could improve the performance of the recipes.

Note that this will first require an update to the `MethodMatcher` implementation to improve its performance when matching parameter types.

### DIFF
--- a/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
@@ -19,6 +19,7 @@ import com.sun.source.tree.AnnotationTree;
 import com.sun.source.tree.MethodTree;
 import com.sun.source.tree.Tree;
 import com.sun.source.tree.VariableTree;
+import com.sun.tools.javac.code.Flags;
 import com.sun.tools.javac.code.Symbol;
 import com.sun.tools.javac.code.Type;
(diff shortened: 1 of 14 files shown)


src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
```diff
@@ -539,9 +540,29 @@ private String generatePreconditions(List<TemplateDescriptor> beforeTemplates, i
...
+                            if (method.name.toString().equals("<init>")) {
+                                sb.append("<constructor>");
+                            } else {
+                                sb.append(method.name.toString());
+                            }
+                            com.sun.tools.javac.util.List<Type> parameterTypes = method.type.getParameterTypes();
+                            if (parameterTypes.isEmpty()) {
+                                sb.append("()");
+                            } else {
+                                if ((method.flags() & Flags.VARARGS) == 0 && parameterTypes.stream().allMatch(t -> t instanceof Type.JCPrimitiveType || t instanceof Type.ClassType)) {
```
knutwannheden, line 556: There should probably also be a predicate here that checks that the method indeed has overloads.

timtebeek:
Given the amount of changes & conflicts since I propose we close this PR, and potentially start anew if we want to add this still.

PR #78, opened 2024-03-08
Link to Picnic's Error Prone Support documentation based on OnlineDocumentation

timtebeek:
As discussed on #47 , this now links the external docs for additional context, as that shows a diff of before and after that might be helpful for users to go over as well, and gives credit there.

cc @rickie

### DIFF
--- a/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
@@ -444,7 +444,8 @@ public class RefasterTemplateProcessor extends TypeAwareProcessor {
 
                 // Extract from the RecipeDescriptor annotation
                 for (JCTree.JCAnnotation annotation : classDecl.getModifiers().getAnnotations()) {
-                    if (annotation.type.toString().equals("org.openrewrite.java.template.RecipeDescriptor")) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
```diff
@@ -466,6 +467,10 @@ private String recipeDescriptor(JCTree.JCClassDecl classDecl, String defaultDisp
                             }
                         }
                         break;
+                    } else if ("tech.picnic.errorprone.refaster.annotation.OnlineDocumentation".equals(annotationFqn)) {
+                        if (annotation.getArguments().isEmpty()) {
```
timtebeek, line 471: It seems rare / unused to have any value here ; they mostly rely on [the default value](https://github.com/PicnicSupermarket/error-prone-support/blob/8f64489fa0dded9255d20c227d0810af1cf52493/refaster-support/src/main/java/tech/picnic/errorprone/refaster/annotation/OnlineDocumentation.java#L44-L48), which isn't immediately easily accessible here for as much as I can tell. Hence why I went for the easiest approach to have some links to the docs already, which we can refine later.

knutwannheden (approved):

rickie:
Amazing 🚀 !

PR #91, opened 2024-06-15
Troubleshoot mismatched AssertJ `AbstractStringAssert.isEqualTo("")`

timtebeek:
This is a recipe that originates with Picnic: https://error-prone.picnic.tech/refasterrules/AssertJStringRules/#abstractstringassertstringisempty

At present it fails to convert this test case
```java
@Test
void convertsIsEqualToEmptyString() {
    rewriteRun(
      //language=java
      java(
        """
          import static org.assertj.core.api.Assertions.assertThat;
          class Test {
              void test() {
                  assertThat("test").isEqualTo("");
              }
          }
          """,
        """
          import static org.assertj.core.api.Assertions.assertThat;
          class Test {
              void test() {
                  assertThat("test").isEmpty();
              }
          }
          """
      )
    );
}
```

As discovered on
- https://github.com/openrewrite/rewrite-testing-frameworks/pull/527

Not yet sure if the fix should go into the generated recipe, or in the matchers used in the generated recipe. Figured log this as a pseudo issue to more easily see the generated recipe.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -93,6 +93,7 @@ dependencies {
     testImplementation("org.junit.jupiter:junit-jupiter-api:latest.release")
     testImplementation("org.junit.jupiter:junit-jupiter-params:latest.release")
     testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:latest.release")
+    testImplementation("org.assertj:assertj-core:latest.release")
 }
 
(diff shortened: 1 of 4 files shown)


timtebeek:
Continued here; as the problem looks to be with the matcher, not the generator.
- https://github.com/openrewrite/rewrite/issues/4423

PR #100, opened 2024-07-31
Match method overrides in precondition

timtebeek:
## What's changed?
Instead of `new UsesMethod<>("java.lang.String join(..)")`
Generate `new UsesMethod<>("java.lang.String join(..)", true)`

## What's your motivation?
We want to also match methods overrridden in sub classes where possible.

Not sure if this might negatively affect some existing recipes that have come to expect stricter matching.
Figured open this PR to get the discussion around that going.

Also not yet sure if the `before.matcher(getCursor())).find()` would even match.

@knutwannheden 

We could add explicit recipes for overridden methods, but that quickly explodes the number of recipes.

We could also offer a way to tag before templates where overridden methods should also be matched.

## Any additional context
Came up while discussing recipes for Apache POI, where `Cell` might have multiple extending classes.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
     signing
 
     id("nebula.maven-resolved-dependencies") version "17.3.2"
-    id("nebula.release") version "15.3.1"
+    id("com.netflix.nebula.release") version "19.0.10"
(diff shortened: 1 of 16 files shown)


knutwannheden (approved):
LGTM.

PR #108, opened 2024-08-26
Verify equality of template signatures

### LINKED ISSUE #107: Mismatch in parameters of before/after not recognized

Due to copy/pasting an existing recipe first I accidentally created an after method with more arguments than the before method:
```java
    static final class Reproducer {
        @BeforeTemplate
        BufferedReader before(Path path, Charset charset) throws IOException {
            return new BufferedReader(
                    new InputStreamReader(new BufferedInputStream(Files.newInputStream(path)), charset));
        }

        @AfterTemplate
        BufferedReader after(Path path, String content, Charset charset) throws IOException {
            // the second argument "content" does not exist in the before template
            return Files.newBufferedReader(path, charset);
        }
    }
```

## What did you expect to see?
It would be good to raise an explicit error that the `@AfterTemplate` method doesn't have the same number (and types and order...) of arguments like the `@BeforeTemplate` method.

## What is the full stack trace of any errors you encountered?
The generated code doesn't compile, therefore the error doesn't go unnoticed. But users might still be confused about the exact reason.
```
FilesRulesRecipes.java:[480,131] incompatible types: <nulltype> cannot be converted to int
```
grafik

Yes.

Side note: The above recipe (and multiple variants of it for other nestings of readers/streams) would be a valid recipe for rewrite-java probably. I just stumbled over this in real code yesterday.

Bananeweizen:
Make sure that all templates belonging to a recipe have the same number, order and type for all the method arguments. This avoids accidental copy/paste errors which otherwise can lead to compiling recipe code, but wrong results at runtime.

- Fixes #107.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
@@ -735,11 +735,18 @@ public class RefasterTemplateProcessor extends TypeAwareProcessor {
             // resolve so that we can inspect the template body
             boolean valid = resolve();
             if (valid) {
+                Set<String> signatures = new HashSet<>();
                 for (TemplateDescriptor template : beforeTemplates) {
                     valid &= template.validate();
+                    signatures.add(template.method.getParameters().stream().map(p -> p.getType().toString()).collect(Collectors.joining(",")));
                 }
                 if (afterTemplate != null) {
                     valid &= afterTemplate.validate();
+                    signatures.add(afterTemplate.method.getParameters().stream().map(p -> p.getType().toString()).collect(Collectors.joining(",")));
+                }
+                if (signatures.size() > 1) {
+                    printNoteOnce("@Before and @After methods of a single recipe must have the same method signature", classDecl.sym);
+                    valid = false;
                 }
             }
             return valid ? this : null;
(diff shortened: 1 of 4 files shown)


Stephan202:
Heads up: this may quite significantly reduce the number of [Error Prone Support Refaster rules](https://github.com/PicnicSupermarket/error-prone-support/tree/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules) for which a corresponding OpenRewrite recipe is generated.

Stock Refaster requires that `@AfterTemplate` method parameter names are also present for each `@BeforeTemplate` method, but not vice versa, and does not require that the types match. We rely on latter to keep rules compact, while the former is valid in case a rule drops an unused subexpression.

Bananeweizen:
Then the only thing remaining to raise an error for would be after.args.size() > before.args.size(). However, that would have found just 1 out of the 3 or 4 copy/paste errors I made in recent weeks that all motivated me to have the strict check. :/

Stephan202:
Yeah, that's painful :(. I recognize that it's _soo_ easy to make copy-paste errors when creating a bunch of similar Refaster rules.

This motivated us to write a test framework (where each rule has before- and after-example code), but as-is said framework isn't maximally user friendly, requiring changes in three files (example: PicnicSupermarket/error-prone-support#1290). At some point I looked into creating a "DSL" that would have example code be "nested" inside the relevant Refaster rule. Something to return to, one of these days :)

knutwannheden:
It seems that we can at least also verify that the parameters of the after templates correspond to a parameter in each before template, where we perform the check by name.

timtebeek:
There have been some parallel efforts lately that have also looked at arguments and their order
- https://github.com/openrewrite/rewrite-templating/pull/127

With that I think we can close this effort;Thanks again all!

PR #111, opened 2024-09-15
Support Refaster templates which use lambdas

knutwannheden:
This still doesn't allow a template to directly return a lambda, because the `JavaTemplate` doesn't capture the lambda's type in any way, but at least a template can still use a lambda (e.g. as a method argument).

The actual fix is to make sure that variable references aren't printed as fully qualified names with the owning method as qualifier.

### DIFF
--- a/src/main/java/org/openrewrite/java/template/internal/TemplateCode.java
+++ b/src/main/java/org/openrewrite/java/template/internal/TemplateCode.java
@@ -146,7 +146,20 @@ public class TemplateCode {
                         imports.add(sym.getQualifiedName().toString());
                     }
                 }
-            } else if (sym instanceof Symbol.MethodSymbol || sym instanceof Symbol.VarSymbol) {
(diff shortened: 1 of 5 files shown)


timtebeek (approved):

timtebeek:
This change brings the totals of rules from ErroProne Support not supported to
```
@Matches is currently not supported=1
@NotMatches is currently not supported=2
Lambdas are currently not supported=2
Method references are currently not supported=3
If statements are currently not supported=8
@Placeholder is currently not supported=15
@Repeated is currently not supported=21
Generic type parameters are currently not supported=519
```

knutwannheden:
I doubt this change will have moved the needle by a lot. Using lambdas as a method argument in a Refaster template is probably fairly uncommon. Also, I don't even think we had a detection for this, only the detection for a template returning a lambda, which unfortunately still isn't supported. We should try to overhaul JavaTemplate so that we can support more Refaster templates.

PR #117, opened 2024-11-23
Visit multiple elements for diverse before templates

timtebeek:
## What's your motivation?
Before this change the order of before templates determined which visit method (singular) to generate, and searched for all before templates from that single visit method. This is not correct, and leads to some before templates not matching when they would require to visit a different element.

Not compile refaster templates that would require us to visit different elements.

## Any additional context
- Fixes https://github.com/moderneinc/rewrite-recipe-starter/issues/66

### DIFF
--- a/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
@@ -353,82 +353,95 @@ public class RefasterTemplateProcessor extends TypeAwareProcessor {
                 }
                 visitor.append("\n");
 
-                List<String> lstTypes = LST_TYPE_MAP.get(getType(descriptor.beforeTemplates.get(0).method));
-                for (String lstType : lstTypes) {
-                    String methodSuffix = lstType.startsWith("J.") ? lstType.substring(2) : lstType;
-                    visitor.append("            @Override\n");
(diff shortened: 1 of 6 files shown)


timtebeek:
Confirmed that this recipe now makes the tests in rewrite-recipe-starter pass
```java
import com.google.errorprone.refaster.annotation.AfterTemplate;
import org.openrewrite.java.template.RecipeDescriptor;

@RecipeDescriptor(
        name = "Standardize empty String checks",
        description = "Replace calls to `String.length() == 0` with `String.isEmpty()`."
)
public class StringIsEmpty {

    @BeforeTemplate
    boolean lengthIsZero(String s) {
        return s.length() == 0;
    }

    @BeforeTemplate
    boolean zeroIsLength(String s) {
        return 0 == s.length();
    }

    @BeforeTemplate
    boolean equalsEmptyString(String s) {
        return s.equals("");
    }

    @BeforeTemplate
    boolean emptyStringEquals(String s) {
        return "".equals(s);
    }

    @BeforeTemplate
    boolean lessThanOne(String s) {
        return s.length() < 1;
    }

    @BeforeTemplate
    boolean oneGreaterThanLength(String s) {
        return 1 > s.length();
    }

    @AfterTemplate
    boolean isEmpty(String s) {
        return s.isEmpty();
    }
}
```

PR #120, opened 2024-12-11
Improve preconditions check

jevanlingen:
## What's changed?
- More or less preconditions in recipes, depending on the used types in the `@BeforeTemplate` methods. Every @BeforeTemplate rules will be its own block of rules, separated by a `Preconditions.or` block. Possible duplicates will be detected and removed of the final precondition set (called _pruning_ of the preconditions).
- Preconditions rules are printed alphabetically (but `UsesType` comes before `UsesMethod`).

## What's your motivation?
More support for picnics refasten templates.

### Checklist


src/test/resources/refaster/PreConditionsVerifier.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.Map;
+
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+
+/**
+ * A refaster template to test when a `UsesType`and Preconditions.or should or should not be applied to the Preconditions check.
+ */
+public class PreConditionsVerifier {
+    public static class NoUsesTypeWhenBeforeTemplateContainsPrimitiveOrString {
```
jevanlingen, line 28: I choose as class name the intention of the recipe, like we use to do for the method name of a @Test method.. This make it both short and easy to grasp.

src/test/resources/refaster/PreConditionsVerifier.java
```diff
@@ -0,0 +1,112 @@
...
+        @BeforeTemplate
+        void before(double actual, int ignore) {
+            System.out.println(actual);
```
jevanlingen, line 31: I kept the `System.out.println` here, which will be translated to `new UsesMethod<>("java.io.PrintStream println(..)", true),`. Let's focus on this PR just on the first improvement of the UsesType.

src/test/resources/refaster/PreConditionsVerifierRecipes.java
```diff
@@ -0,0 +1,403 @@
...
+                                ctx,
+                                SHORTEN_NAMES
+                        );
+                    }
+                    return super.visitMethodInvocation(elem, ctx);
+                }
+
+            };
+            return Preconditions.check(
+                    new UsesMethod<>("java.io.PrintStream println(..)", true),
```
jevanlingen, line 114: Name: **NoUsesTypeWhenBeforeTemplateContainsPrimitiveOrStringRecipe**, thus:
- no `new UsesType<>("java.lang.Double", true)`
- no `new UsesType<>("java.util.Map", true)`

src/test/resources/refaster/PreConditionsVerifierRecipes.java
```diff
@@ -0,0 +1,403 @@
...
+                                SHORTEN_NAMES
+                        );
+                    }
+                    return super.visitMethodInvocation(elem, ctx);
+                }
+
+            };
+            return Preconditions.check(
+                    Preconditions.and(
+                            new UsesType<>("java.util.Map", true),
```
jevanlingen, line 182: Name: **NoUsesTypeWhenBeforeTemplateContainsPrimitiveAndAnotherTypeRecipe**, thus:
- no `new UsesType<>("java.util.Map", true)`

src/test/resources/refaster/PreConditionsVerifierRecipes.java
```diff
@@ -0,0 +1,403 @@
...
```
jevanlingen, line 252: Name: **NoUsesTypeWhenBeforeTemplateContainsStringAndAnotherTypeRecipe**, thus:
- no `new UsesType<>("java.util.Map", true)`

src/test/resources/refaster/PreConditionsVerifierRecipes.java
```diff
@@ -0,0 +1,403 @@
...
```
jevanlingen, line 321: Name: **UsesTypeMapWhenBeforeTemplateContainsMapOnlyRecipe**, thus:
- Use `new UsesType<>("java.util.Map", true)`

src/test/resources/refaster/PreConditionsVerifierRecipes.java
```diff
@@ -0,0 +1,403 @@
...
+                        );
+                    }
+                    return super.visitMethodInvocation(elem, ctx);
+                }
+
+            };
+            return Preconditions.check(
+                    Preconditions.and(
+                            new UsesMethod<>("java.io.PrintStream println(..)", true),
+                            Preconditions.or(
```
jevanlingen, line 393: Name: **UsesTypeMapOrListWhenBeforeTemplateContainsMapAndListRecipe**, thus:
- Use `new UsesType<>("java.util.List", true)` OR `new UsesType<>("java.util.Map", true)`

src/test/resources/refaster/PreConditionsVerifierRecipes.java
```diff
@@ -0,0 +1,403 @@
...
+                    if ((matcher = before.matcher(getCursor())).find()) {
+                        return embed(
+                                after.apply(getCursor(), elem.getCoordinates().replace(), matcher.parameter(0)),
+                                getCursor(),
+                                ctx,
+                                SHORTEN_NAMES
+                        );
+                    }
+                    if ((matcher = before0.matcher(getCursor())).find()) {
+                        maybeRemoveImport("java.util.Map");
```
jevanlingen, line 168: Should this stay? I think so, right? _If I am correct, the `maybeRemoveImport` is smart enough to figure out whether the import should be removed or not._

timtebeek: Indeed; imports will only be removed if there are no more LST elements using those types. In addition to that the `before0.matcher` already indicated we're only handling cases where there is a Map being matched.

src/test/resources/refaster/PreConditionsVerifierRecipes.java
```diff
@@ -0,0 +1,403 @@
...
+                    }
+                    return super.visitMethodInvocation(elem, ctx);
+                }
+
+            };
+            return Preconditions.check(
+                    Preconditions.and(
+                            new UsesType<>("java.util.Map", true),
+                            new UsesMethod<>("java.io.PrintStream println(..)", true)
+                    ),
```
timtebeek, line 184:
```suggestion
                    new UsesMethod<>("java.io.PrintStream println(..)", true),
```

src/test/resources/refaster/PreConditionsVerifierRecipes.java
timtebeek: Turned into a code suggestion just below such that there's no ambiguity, as the wrapping `and` can also be removed.

src/test/resources/refaster/PreConditionsVerifierRecipes.java
```diff
@@ -0,0 +1,403 @@
...
```
timtebeek, line 254:
```suggestion
                    new UsesMethod<>("java.io.PrintStream println(..)", true),
```

src/test/resources/refaster/PreConditionsVerifierRecipes.java
timtebeek: In this case indeed I would not expect any change to the preconditions.

src/test/resources/refaster/PreConditionsVerifierRecipes.java
timtebeek: In this case I'd expect no change to the preconditions either, as we already have that `or` between the method and types in use.

timtebeek (commented):
Great to see the tests already; I've placed some code suggestions to rule out any chance of miscommunication. Thanks for taking this on!

src/test/resources/refaster/PreConditionsVerifier.java
jevanlingen: Notice this may leads to o performance loss. Imagine a recipe looking like:

```
import com.sun.tools.javac.util.Convert;

public class SimpleTest {
        @BeforeTemplate
        void before(double actual, String value) {
            Convert.quote(value);
            System.out.println(actual);
        }

        @BeforeTemplate
        void beforeTwo(String actual, String ignore) {
            Convert.quote(value);
            System.out.println(actual);
        }

        @AfterTemplate
        void after(Object actual) {
            System.out.println("Changed: " + actual);
        }
    }
 ```   

Currently the code will add a `new UsesType<>("com.sun.tools.javac.util.Convert", true),`. In the most naive new version, we can skip the UsesMethod creation if:
There is more than one `@BeforeTemplate` AND one of the arguments contain a primitive.

So maybe it would be good to still add one of those test and also take above scenario into account.

File: src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java [hunk 11]
```diff
@@ -688,7 +688,13 @@ public void scan(JCTree jcTree) {
...
                             ((Symbol.MethodSymbol) jcIdent.sym.owner).params.contains(jcIdent.sym) &&
                             seenParams.add(jcIdent.sym)) {
-                            afterParams.add(beforeParamOrder.get(((Symbol.MethodSymbol) jcIdent.sym.owner).params.indexOf(jcIdent.sym)));
+                            Integer idx = beforeParamOrder.get(((Symbol.MethodSymbol) jcIdent.sym.owner).params.indexOf(jcIdent.sym));
+                            // TODO fix ugly hack
+                            if (idx == null && beforeParamOrder.size() == 1) {
+                                afterParams.add(0);
+                            } else {
+                                afterParams.add(idx);
+                            }
```
timtebeek, line 697: This is likely a result of having before/after templates with different types, and/or different number of arguments, meaning the look up fails. Possibly we'll want to look at the names of the variables used, but that's something to explore.

jevanlingen: Reverted it, I understood now what I did wrong. There is no support for beforeTemplates where you do not use the incoming argument. For example:

```java
@BeforeTemplate
String string(String hello) {
    return "world"'
}
```

But it does not make any sense at all to support such a `@BeforeTemplate`. So I don't think we should ponder about this any longer.

src/test/resources/refaster/PreConditionsVerifierRecipes.java
```diff
@@ -0,0 +1,561 @@
...
+                    }
+                    return super.visitMethodInvocation(elem, ctx);
+                }
+
+            };
+            return Preconditions.check(
+                    Preconditions.and(
+                            new UsesType<>("com.sun.tools.javac.util.Convert", true),
+                            new UsesMethod<>("com.sun.tools.javac.util.Convert quote(..)", true),
+                            new UsesMethod<>("java.lang.String valueOf(..)", true)
```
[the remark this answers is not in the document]
jevanlingen: You are right. For current solution, this recipe is not pruned (because the arguments of the beforeTemplates are all primitives). So this is the 'normal' or/and-type-method-logic. I will change this piece of code to what should happen, tomorrow I'll take a look at the implementation.

src/main/java/org/openrewrite/java/template/processor/PreCondition.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private static String joinPreconditions(Collection<PreCondition> rules, String op, int indent) {
+        if (rules.isEmpty()) {
+            return "";
+        } else if (rules.size() == 1) {
+            return rules.iterator().next().toString();
+        }
+        char[] indentChars = new char[indent];
+        Arrays.fill(indentChars, ' ');
+        String indentStr = new String(indentChars);
```
timtebeek, line 54: String.format also has a way to add left/right padding with `%4s` for instance; if you replace the string concatenation below then that might help readability and remove the need for these repeated strings.

jevanlingen: Hey, I just copied this function 😉

timtebeek:
As a quick note, this likely heavily conflicts with this earlier effort; could help to explore the recipe that was added there:
- https://github.com/openrewrite/rewrite-templating/pull/77

src/main/java/org/openrewrite/java/template/processor/PreCondition.java
jevanlingen: I don't think it get much clearer when using things like `%4s`. I think only be much unclearer if you write it like:

```java
String.format("Preconditions.%s(\n%" + indent + "s%s" + indentStr.substring(0, indent - 4) + ')',
                op, String.join(",\n" + indentStr, preconditions));
```

src/main/java/org/openrewrite/java/template/processor/PreCondition.java
```diff
@@ -0,0 +1,120 @@
...
+    @Value
+    @EqualsAndHashCode(callSuper = false)
+    @RequiredArgsConstructor
+    public static class Or extends PreCondition {
+        Set<PreCondition> preConditions;
+        int indent;
+
+        @Override
+        boolean fitsInto(PreCondition p) {
+            throw new NotImplementedException();
```
jevanlingen, line 54: Improve..

src/main/java/org/openrewrite/java/template/processor/PreCondition.java
```diff
@@ -0,0 +1,120 @@
...
+                return preConditions.contains(p);
+            } else if (p instanceof Or) {
+                throw new NotImplementedException();
+            } else if (p instanceof And) {
+                if (preConditions.size() > ((And) p).preConditions.size()) {
+                    return false;
+                }
+                return preConditions.stream().allMatch(it -> it.fitsInto(p));
+            }
+            throw new IllegalArgumentException("Type is not supported: " + p.getClass());
```
jevanlingen, line 99: @timtebeek This branch is never possible (I would have modelled a sealed hierarchy if we would have used a newer Java version). Would just `return false` be ok for this case?

timtebeek: Yes I think that'd be preferred, if `return false` leads to not making any changes.

src/main/java/org/openrewrite/java/template/processor/Precondition.java
```diff
@@ -0,0 +1,135 @@
...
+    @Value
+    @EqualsAndHashCode(callSuper = false)
+    @RequiredArgsConstructor
+    public static class Or extends Precondition {
+        Set<Precondition> preConditions;
+        int indent;
+
+        @Override
+        boolean fitsInto(Precondition p) {
+            throw new NotImplementedException();
```
Laurens-W, line 69: Should this be implemented?

src/main/java/org/openrewrite/java/template/processor/Precondition.java
```diff
@@ -0,0 +1,135 @@
...
+    public static class And extends Precondition {
+        Set<Precondition> preConditions;
+        int indent;
+
+        @Override
+        boolean fitsInto(Precondition p) {
+            if (p instanceof Rule) {
+                return preConditions.contains(p);
+            } else if (p instanceof Or) {
+                throw new NotImplementedException();
```
Laurens-W, line 107: same here

src/main/java/org/openrewrite/java/template/processor/Precondition.java
jevanlingen: yes, comment is lost by renaming of file...

src/test/resources/refaster/NestedPreconditionsRecipe.java
```diff
@@ -96,17 +96,16 @@ public J visitNewClass(J.NewClass elem, ExecutionContext ctx) {
 
         };
         return Preconditions.check(
```
Laurens-W, line 98: I'm just being really nitpicky but wouldn't it be great if this would be a check of `and(common, or(and(hashmap related), and(linkedhashmap related))`?

src/test/resources/refaster/RefasterAnyOfRecipes.java
```diff
@@ -192,17 +192,16 @@ public J visitMethodInvocation(J.MethodInvocation elem, ExecutionContext ctx) {
 
             };
             return Preconditions.check(
-                    Preconditions.and(
```
Laurens-W, line 195: Same here

src/test/java/org/openrewrite/java/template/processor/PreconditionTest.java
```diff
@@ -0,0 +1,119 @@
...
+ */
+package org.openrewrite.java.template.processor;
+
+import org.junit.jupiter.api.Test;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+import static com.google.common.truth.Truth.assertThat;
```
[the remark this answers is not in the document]
jevanlingen: On purpose, assertJ is not on the class path. We use [Truth](https://truth.dev) instead in this repo. Don't ask me why 😉.

src/test/resources/refaster/PreConditionsVerifier.java
```diff
@@ -0,0 +1,147 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package foo;
+
+import java.util.List;
+import java.util.Map;
+
```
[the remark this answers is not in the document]
timtebeek: The bot is a little off again; here's what the patch looks like
```diff
diff --git a/src/test/resources/refaster/PreConditionsVerifier.java b/src/test/resources/refaster/PreConditionsVerifier.java
index ceb8bb3..1f749cc 100644
--- a/src/test/resources/refaster/PreConditionsVerifier.java
+++ b/src/test/resources/refaster/PreConditionsVerifier.java
@@ -15,13 +15,13 @@
  */
 package foo;
 
-import java.util.List;
-import java.util.Map;
-
 import com.google.errorprone.refaster.annotation.BeforeTemplate;
 import com.sun.tools.javac.util.Constants;
 import com.sun.tools.javac.util.Convert;
 
+import java.util.List;
+import java.util.Map;
+
```

src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
```diff
@@ -607,48 +606,25 @@ private Set<String> getImportsAsStrings(Map<TemplateDescriptor, Set<String>> imp
...
-                if (preconditions.size() == 1) {
-                    return joinPreconditions(preconditions.values().iterator().next(), "and", indent + 4);
-                } else if (preconditions.size() > 1) {
-                    Set<String> common = new LinkedHashSet<>();
-                    for (String dep : preconditions.values().iterator().next()) {
-                        if (preconditions.values().stream().allMatch(v -> v.contains(dep))) {
-                            common.add(dep);
+                        if (!usesVisitors.isEmpty()) {
+                            preconditions.put(beforeTemplate.method.name.toString() + (arity == 1 ? "" : "$" + i), usesVisitors);
                         }
```
timtebeek, line 614: Should this instead be 
```suggestion
                        if (!usesVisitors.isEmpty()) {
                            preconditions.put(beforeTemplate.method.name.toString() + (arity == 1 ? "" : "$" + i), usesVisitors);
                        } else {
                            return null; // At least one of the before templates has no preconditions, so we can not use any preconditions
                        }
```

jevanlingen: You are right; my test did not caught this problem, because there was alway the `java.io.PrintStream` type... For example:

```java
@BeforeTemplate
void _int(int actual) {
  System.out.println(actual);
}
```        

I will change the tests accordingly!

src/test/resources/refaster/NestedPreconditionsRecipe.java
jevanlingen: Done!

src/test/resources/refaster/MethodThrowsRecipe.java
```diff
@@ -85,8 +85,8 @@ public J visitMethodInvocation(J.MethodInvocation elem, ExecutionContext ctx) {
         };
         return Preconditions.check(
                 Preconditions.and(
-                        new UsesType<>("java.nio.file.Files", true),
                         new UsesType<>("java.nio.charset.StandardCharsets", true),
+                        new UsesType<>("java.nio.file.Files", true),
```
knutwannheden, line 89: Not directly related to the changes of this PR, but for `UsesType` I am thinking we can probably set the `includeImplicit` parameter to `false`. What it does is to check the declaring type, return type, and parameter types of each used method, whereas `UsesType` otherwise basically only checks if the source code has some variable or expression of the given type.

It is rather difficult to reason about, so maybe it is safest to leave it as is.

jevanlingen: Agreed, let's consider this at a later stage.

timtebeek (approved):
Great to have the proper minimized preconditions generated now. Thanks a lot!

PR #124, opened 2024-12-23
Tweak how JavaDoc is converted into displayName/description

timtebeek:
## What's changed?
Now set either description or displayName + description, but drop the longer generated description where possibe.

## What's your motivation?
Previously we would often retain the generated Markdown that merely showed the input template, whereas there's JavaDoc comments that would be a better fit and make the docs more readable.

### DIFF
--- a/src/test/resources/refaster/PicnicRulesRecipes.java
+++ b/src/test/resources/refaster/PicnicRulesRecipes.java
@@ -61,7 +61,8 @@ public class PicnicRulesRecipes extends Recipe {
                 new FirstRuleRecipe(),
                 new SecondRuleRecipe(),
                 new ThirdRuleRecipe(),
-                new FourthRuleRecipe()
+                new FourthRuleRecipe(),
+                new FifthRuleRecipe()
         );
     }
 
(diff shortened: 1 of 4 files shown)


src/test/resources/refaster/PicnicRulesRecipes.java
```diff
@@ -52,37 +52,205 @@ public String getDisplayName() {
...
         @Override
         public String getDisplayName() {
-            return "Refaster template `PicnicRules.NestedRule`";
+            return "Refaster template `PicnicRules.FirstRule`";
         }
 
         @Override
         public String getDescription() {
-            return "Recipe created for the following Refaster template:\n```java\npublic static class NestedRule {\n    \n    @BeforeTemplate()\n    String before(String s, String s1, String s2) {\n        return s.replaceAll(s1, s2);\n    }\n    \n    @AfterTemplate()\n    String after(String s, String s1, String s2) {\n        return s != null ? s.replaceAll(s1, s2) : s;\n    }\n}\n```\n.";
+            return "A single line used as description";
```
timtebeek, line 88: This generated block greatly inflates our docs, which hurts readability. Better then to use the JavaDoc provided instead.

jevanlingen (approved):

Issue #126, opened 2025-03-03
@Matches not matching correct parameters


fmodesto:
It seems the generated code for @Matches doesn't respect the order of the parameters. It will generate code for the current position @Matches is in the parameter list, instead of the position is going to be in the template.

## What version of OpenRewrite are you using?
Latest BOM 3.3.0

## How are you running OpenRewrite?
I'm using  rewrite-recipe-starter

## What is the smallest, simplest way to reproduce the problem?
```java
public class FlipEquals {

    static class Literal implements Matcher<Expression> {
        @Override
        public boolean matches(Expression expression) {
            return expression instanceof J.Literal;
        }
    }

    @BeforeTemplate
    boolean before(@Matches(Literal.class) String literal, @NotMatches(Literal.class) String str) {
        return str.equals(literal);
    }

    @AfterTemplate
    boolean after(String literal, String str) {
        return literal.equals(str);
    }
}
```

## What did you expect to see?

```java
"literal".equals(str)
```

## What did you see instead?

```java
str.equals("literal")
```

## The generated code
```java
            final JavaTemplate before = JavaTemplate
                    .builder("#{str:any(java.lang.String)}.equals(#{literal:any(java.lang.String)})")
                    .build();
            final JavaTemplate after = JavaTemplate
                    .builder("#{literal:any(java.lang.String)}.equals(#{str:any(java.lang.String)})")
                    .build();

            @Override
            public J visitMethodInvocation(J.MethodInvocation elem, ExecutionContext ctx) {
                JavaTemplate.Matcher matcher;
                if ((matcher = before.matcher(getCursor())).find()) {
                    if (!new com.yourorg.FlipEquals.Literal().matches((Expression) matcher.parameter(0))) {
                        return super.visitMethodInvocation(elem, ctx);
                    }
                    if (new com.yourorg.FlipEquals.Literal().matches((Expression) matcher.parameter(1))) {
                        return super.visitMethodInvocation(elem, ctx);
                    }
                    return embed(
                            after.apply(getCursor(), elem.getCoordinates().replace(), matcher.parameter(1), matcher.parameter(0)),
                            getCursor(),
                            ctx,
                            SHORTEN_NAMES, SIMPLIFY_BOOLEANS
                    );
                }
                return super.visitMethodInvocation(elem, ctx);
            }
```

matcher.parameter(0) or `str` is match for Literal
matcher.parameter(1) or `literal` is not match for literal.

timtebeek:
[...]

PR #127, opened 2025-03-03
Fix issue #126 @Matches applies to the wrong parameter

fmodesto:
## What's changed?
- Fix issue #126.

Use the argument name to match parameters in matcher.

## What's your motivation?
Current implementation is a bit indeterministic.  It relies in the parameters to be in the same order as expressed in the body to work reliable. With this merge, a change in the order of the parameters won't affect the replacements.

This might break current recipes.  See fixes in unrelated tests.
Should we do a validation for correctness? Checking that all the parameter names in @AfterTemplate are found in all @BeforeTemplate https://github.com/openrewrite/rewrite-templating/pull/108#issuecomment-2310842563

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
@@ -390,20 +390,20 @@ public class RefasterTemplateProcessor extends TypeAwareProcessor {
                     for (int i = 0; i < arity; i++) {
                         visitMethod.append("                if (" + "(matcher = ").append(entry.getKey()).append(arity > 1 ? "$" + i : "").append(".matcher(getCursor())).find()").append(") {\n");
                         com.sun.tools.javac.util.List<JCTree.JCVariableDecl> jcVariableDecls = entry.getValue().method.getParameters();
-                        for (int j = 0; j < jcVariableDecls.size(); j++) {
-                            JCTree.JCVariableDecl param = jcVariableDecls.get(j);
-                            int index = beforeParameters.getOrDefault(param.name, -1);
+                        for (JCTree.JCVariableDecl param : jcVariableDecls) {
                             com.sun.tools.javac.util.List<JCTree.JCAnnotation> annotations = param.getModifiers().getAnnotations();
                             for (JCTree.JCAnnotation jcAnnotation : annotations) {
                                 String annotationType = jcAnnotation.attribute.type.tsym.getQualifiedName().toString();
-                                if (annotationType.equals("org.openrewrite.java.template.NotMatches")) {
+                                if (!beforeParameters.containsKey(param.name) && annotationType.startsWith("org.openrewrite.java.template")) {
+                                    printNoteOnce("Ignoring annotation " + annotationType + " on unused parameter " + param.name, entry.getValue().classDecl.sym);
+                                } else if (annotationType.equals("org.openrewrite.java.template.NotMatches")) {
                                     String matcher = ((Type.ClassType) jcAnnotation.attribute.getValue().values.get(0).snd.getValue()).tsym.getQualifiedName().toString();
-                                    visitMethod.append("                    if (new ").append(matcher).append("().matches((Expression) matcher.parameter(").append(index).append("))) {\n");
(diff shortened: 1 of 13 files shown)


src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
```diff
@@ -383,23 +385,25 @@ private String generateVisitMethod(Map<String, TemplateDescriptor> beforeTemplat
...
                 visitMethod.append("                JavaTemplate.Matcher matcher;\n");
                 for (Map.Entry<String, TemplateDescriptor> entry : beforeTemplates.entrySet()) {
+                    Map<Name, Integer> beforeParameters = findParameterOrder(entry.getValue().method);
                     int arity = entry.getValue().getArity();
                     for (int i = 0; i < arity; i++) {
                         visitMethod.append("                if (" + "(matcher = ").append(entry.getKey()).append(arity > 1 ? "$" + i : "").append(".matcher(getCursor())).find()").append(") {\n");
                         com.sun.tools.javac.util.List<JCTree.JCVariableDecl> jcVariableDecls = entry.getValue().method.getParameters();
                         for (int j = 0; j < jcVariableDecls.size(); j++) {
                             JCTree.JCVariableDecl param = jcVariableDecls.get(j);
+                            int index = beforeParameters.getOrDefault(param.name, -1);
```
fmodesto, line 395: This happens when the @BeforeTemplate declares an argument, but doesn't use it. Should this be an error? or let it be and skip the logic for @Matches/@NotMatches?

src/test/resources/refaster/MatchingRecipes.java
```diff
@@ -112,7 +112,7 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
                 public J visitMethodInvocation(J.MethodInvocation elem, ExecutionContext ctx) {
                     JavaTemplate.Matcher matcher;
                     if ((matcher = before.matcher(getCursor())).find()) {
-                        if (new org.openrewrite.java.template.MethodInvocationMatcher().matches((Expression) matcher.parameter(1))) {
+                        if (new org.openrewrite.java.template.MethodInvocationMatcher().matches((Expression) matcher.parameter(0))) {
```
fmodesto, line 115: 0 is the correct argument
```java
        @BeforeTemplate
        boolean before(int i, @NotMatches(MethodInvocationMatcher.class) String s) {
            return s.substring(i).isEmpty();
        }
```

src/test/resources/refaster/PreconditionsVerifier.java
```diff
@@ -72,8 +72,8 @@ String before(int value) {
         }
 
         @AfterTemplate
-        Object after(Object actual) {
-            return Convert.quote(String.valueOf(actual));
+        Object after(Object value) {
```
fmodesto, line 75: If we do "fuzzy logic" it will detect that all arguments (none) that exist in `@BeforeTemplate` match the positions in `@AfterTemplate`. Then `actual` would be replace by the `value` argument.

timtebeek:
Awesome to see you come through with an immediate fix @fmodesto ! I'm tagging @knutwannheden for a review as well as I'm out.

@rickie out of interest: how consistent are you with error-prone-support in naming method arguments? Are those the same in before/after? Or might those differ?

rickie:
In Error Prone (Support) this is a bit more strict.

All method arguments are named exactly the same. This is required by Refaster and will throw an error if we don't do that in EPS. It's also required that all method arguments are present in _all_ before and after templates with the same names. Otherwise Refaster won't know how to map the code that matches the method arguments.

knutwannheden:
Let's also require the same in our annotation processor then.

fmodesto:
I also run this recipe to check the order of params.
```java
class Template1 {
  @BeforeTemplate int before(int a, int b) { return a + b; }
  @AfterTemplate int after(int a, int b) { return a; }
}
class Template2 {
  @BeforeTemplate int before(int b, int a) { return a + b; } // swap input params
  @AfterTemplate int after(int a, int b) { return a; }
}
```
both produce the same output
```java
class Input {
  public void example(int x, int y) {
    System.out.println(x + y);
  }
}

class Output {
  public void example(int x, int y) {
    System.out.println(x);
  }
}
```

knutwannheden:
This looks very promising! I will review it tomorrow.

knutwannheden (approved):

knutwannheden:
Thanks again for this PR!

PR #128, opened 2025-03-17
Generic type argument

timtebeek:
### Checklist
- [ ] I've used the IntelliJ IDEA auto-formatter on affected files


src/test/resources/refaster/CollectionIsEmptyRecipe.java
```diff
@@ -0,0 +1,118 @@
...
+            final JavaTemplate before$2 = JavaTemplate
+                    .builder("#{collection:any(java.util.Collection<T>)}.size() < 1")
+                    .build();
+                    .build();
+            final JavaTemplate after = JavaTemplate
+                    .builder("#{collection:any(java.util.Collection<T>)}.isEmpty()")
+                    .build();
+
+            @Override
+            public J visitMethodInvocation(J.MethodInvocation elem, ExecutionContext ctx) { // TODO visitBinary
```
timtebeek, line 80: This generated method is wrong; possibly due to the Refaster.anyOf on the input.

src/test/resources/refaster/CollectionStreamFindIsEmpty.java
```diff
@@ -0,0 +1,36 @@
...
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+
+import java.util.Collection;
+
+class CollectionStreamFindIsEmpty<T> {
+    @BeforeTemplate
+    boolean before(Collection<T> collection) {
+        return Refaster.anyOf(
+                collection.stream().findAny().isEmpty(),
+                collection.stream().findFirst().isEmpty());
```
timtebeek, line 29: These currently fail as the tests here compile with Java 8, whereas these methods are from newer Java versions.

src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
```diff
@@ -735,7 +735,7 @@ public RuleDescriptor(JCTree.JCClassDecl classDecl, JCCompilationUnit cu, Contex
 
             if (classDecl.typarams != null && !classDecl.typarams.isEmpty()) {
                 printNoteOnce("Generic type parameters are currently not supported", classDecl.sym);
-                return null;
+                //TODO return null;
```
timtebeek, line 738: We'll need to figure out in which exact cases we can tolerate the generic types and still end up with working recipes.

src/test/resources/refaster/CollectionIsEmptyRecipe.java
```diff
@@ -0,0 +1,117 @@
...
+                    .build();
+            final JavaTemplate before$2 = JavaTemplate
+                    .builder("#{collection:any(java.util.Collection<T>)}.size() < 1")
+                    .build();
+            final JavaTemplate after = JavaTemplate
+                    .builder("#{collection:any(java.util.Collection<T>)}.isEmpty()")
+                    .build();
+
+            @Override
+            public J visitMethodInvocation(J.MethodInvocation elem, ExecutionContext ctx) { // FIXME visitBinary
```
timtebeek, line 79: The generator is wrong here to not use `visitBinary`.

timtebeek:
- This depends on https://github.com/openrewrite/rewrite-templating/pull/129

timtebeek:
@knutwannheden I was wondering if you have any early feedback on this PR; I've checked that the recipes do match when we remove the generic type place holders or wildcards, but I'm not sure if there might be any other negative effects. Do you think this is worth pursuing?

I've also ran this again error-prone-support, but found there's an as of yet unexplained failure there with the current iteration, so we're not quite ready for a review & merge just yet.

src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
knutwannheden: AFAICT we can use this approach for any unbound type parameters (so basically `T` or `T extends Object`), but as soon as there are bounds this would go wrong. But I think this would already represent an improvement.

We would also have to make sure to not generate anything for templates using the type parameter directly for a parameter like `boolean before(Collection<T> c, T t) { return c.contains(t); }`, as I suspect that this would not work properly.

knutwannheden:
> I've also ran this again error-prone-support, but found there's an as of yet unexplained failure there with the current iteration, so we're not quite ready for a review & merge just yet.

Maybe that will be related to your comment I replied to.

I think we can merge this if we constrain it a bit more to the scenarios where it is applicable.

timtebeek:
With the parallel efforts of @fmodesto I think we can close this one for now; the only part worth restoring in the near future is allowing rules where there are generic arguments by removing this conditional:
https://github.com/openrewrite/rewrite-templating/blob/a3633630f347c1c42ba4afe53f5dae92d5482da9/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java#L743-L746

PR #129, opened 2025-03-18
Refaster.anyOf should visitBinary where applicable

timtebeek:
Previously we always used `visitMethodInvocation` when `Refaster.anyOf` was used; that's not right, and leads to various mismatches. The approach seen here is better in that we use the types of the arguments to `Refaster.anyOf`, to handle cases as seen here:
https://github.com/PicnicSupermarket/error-prone-support/blob/bb01447760dd97fc5977b498b9d5b202f0bbb26f/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/CollectionRules.java#L48-L56

Notice how that now ought to generate `visitBinary` as well as `visitMethodInvocation`.
```
    boolean before(Collection<T> collection) {
      return Refaster.anyOf(
          collection.size() == 0,
          collection.size() <= 0,
          collection.size() < 1,
          Iterables.isEmpty(collection),
          collection.stream().findAny().isEmpty(),
          collection.stream().findFirst().isEmpty());
    }
```

### DIFF
--- a/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
+++ b/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
@@ -50,6 +50,7 @@ import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static java.util.Collections.singleton;
 import static java.util.Collections.singletonList;
 import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.*;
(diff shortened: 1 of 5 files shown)


timtebeek:
The diff here is best viewed when ignoring whitespace, as I've extracted a named RecipeWriter class after repeatedly failing to spot where methods are defined in the IntelliJ sidebar due to heavy use of anonymous classes.

src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java
```diff
@@ -124,528 +126,536 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
...
+                Set<Class<? extends JCTree>> types = entry.getValue().getTypes();
+                for (Class<? extends JCTree> type : types) {
+                    for (String lstType : LST_TYPE_MAP.get(type)) {
+                        templatesByLstType.computeIfAbsent(lstType, k -> new TreeMap<>())
+                                .put(entry.getKey(), entry.getValue());
+                    }
                 }
+            }
+            templatesByLstType.forEach((lstType, typeBeforeTemplates) ->
+                    visitor.append(generateVisitMethod(typeBeforeTemplates, after, descriptor, lstType)));
```
timtebeek, line 380: This is the heart of the changes here: previously we used `groupingBy` with a faulty `getType` method, that instead should support multiple types when it handles `Refaster.anyOf` correctly.

src/test/resources/refaster/RefasterAnyOfRecipes.java
```diff
@@ -190,6 +196,30 @@ public J visitMethodInvocation(J.MethodInvocation elem, ExecutionContext ctx) {
...
+                    if ((matcher = before$0.matcher(getCursor())).find()) {
+                        maybeRemoveImport("java.util.LinkedList");
+                        return embed(
+                                after.apply(getCursor(), elem.getCoordinates().replace()),
+                                getCursor(),
+                                ctx,
+                                SHORTEN_NAMES
+                        );
+                    }
+                    if ((matcher = before$1.matcher(getCursor())).find()) {
```
timtebeek, line 211: In a perfect world we'd only match `before$0` in `visitNewClass`,
and only match `before$1` in `visitMethodInvocation`.

knutwannheden:
Nice find! Will review tomorrow.

knutwannheden (approved):

PR #130, opened 2025-04-05
Fix parameter order when using Refaster.anyOf

fmodesto:
## What's changed?
Fix parameter order when using Refaster.anyOf
Deprecate anyArray()

@knutwannheden 

### Checklist


timtebeek (approved):
Thanks a lot for the detailed work & fix here. Should help fix some of the recipes generated for error-prone-support and further close the gap.

knutwannheden (approved):

knutwannheden:
I did not review the changes in detail, but if all teats pass, it is already a good sign and we can still add more tests if we find regressions.

PR #131, opened 2025-04-06
Fix compiler error on JDK9+

fmodesto:
## What's changed?
Fix rewrite-templating from breaking the java compiler JDK9+

## What's your motivation?
After enable support for generics the compilation starts failing.

The code is mostly extracted from [lombok](https://github.com/projectlombok/lombok/blob/v1.18.36/src/core/lombok/javac/JavacResolution.java#L262). Lombok seems to use some stub for [`ArgumentAttr`](https://github.com/projectlombok/lombok/blob/master/src/stubs/com/sun/tools/javac/comp/ArgumentAttr.java), I just loaded via reflection.

## Any additional context
It seems after running the type resolution, the compiler caches some type information. Later, on the real compilation, it uses this cache and the types mismatch.

This is the minimal template that reproduces the error.
```
package com.yourorg;

import java.util.Optional;

public class Foo<T> {

    @BeforeTemplate
    public T before(Optional<T> a, Optional<T> b) {
        return a.orElseGet(() -> b.get());
    }
}
```

```
[INFO] --- compiler:3.13.0:compile (default-compile) @ rewrite-recipe-starter ---
[INFO] Recompiling the module because of changed source code.
[INFO] Compiling 11 source files with javac [debug target 11] to target\classes
[INFO] Some messages have been simplified; recompile with -Xdiags:verbose to get full output
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] /C:/Users/fran/Projects/rewrite-recipe-starter/src/main/java/com/yourorg/Foo.java:[12,39] incompatible types: bad return type in lambda expression
    T cannot be converted to T
[INFO] 1 error
```

Compilation works with JDK 8 but breaks after that

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/template/internal/JavacResolution.java
+++ b/src/main/java/org/openrewrite/java/template/internal/JavacResolution.java
@@ -254,7 +254,7 @@ public class JavacResolution {
             }
         }
 
-        public static Map<?, ?> enableTempCache(Context context) {
+        public static @Nullable Map<?, ?> enableTempCache(Context context) {
             if (ARGUMENT_TYPE_CACHE == null) {
                 return null;
             }
--- a/src/main/java/org/openrewrite/java/template/internal/JavacResolution.java
+++ b/src/main/java/org/openrewrite/java/template/internal/JavacResolution.java
@@ -34,6 +34,8 @@ import org.jspecify.annotations.Nullable;
 
 import javax.tools.JavaFileObject;
 import java.lang.reflect.Field;
+import java.lang.reflect.Method;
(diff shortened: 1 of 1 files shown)


timtebeek (approved):
Great to see! I'd indeed also stumbled over these same issues in #128; thanks for diving in for the fix. I've also tagged Knut for a review here, as he's more experienced when it comes to compiler internals.

knutwannheden:
Thanks a lot for the fix! So far we hadn't encountered this ourselves yet.

PR #132, opened 2025-04-06
Fix JDK24 compilation error

fmodesto:
## What's changed?
Sync with [lombok code](https://github.com/projectlombok/lombok/blob/master/src/utils/lombok/javac/Javac.java#L187)

## What's your motivation
OpenJDK has remove the `TypeTag.UNKNOWN` constant in JDK24.
Any attempt to use `Javac` class will result in the following exception.
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.13.0:compile (default-compile) on project rewrite-recipe-starter: Fatal error compiling: java.lang.ExceptionInInitializerError: com.sun.tools.javac.code.TypeTag :: UNKNOWN -> [Help 1]
```

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/template/internal/Javac.java
+++ b/src/main/java/org/openrewrite/java/template/internal/Javac.java
@@ -24,6 +24,7 @@ package org.openrewrite.java.template.internal;
 import org.openrewrite.java.template.internal.JavacTreeMaker.TypeTag;
 
 import static org.openrewrite.java.template.internal.JavacTreeMaker.TypeTag.typeTag;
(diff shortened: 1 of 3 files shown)


knutwannheden:
Thanks for fixing this. We've noticed the same problem in `rewrite-java` when compiling with JDK 24. I am thinking we can probably add some reflection to `rewrite-java-21` for this (similar to here). The alternative would be to create a `rewrite-java-24`, but then we should make sure that also works well with 25-ea builds and since 25 will be the new LTS, I think I would personally prefer a `rewrite-java-25`, which then also wouldn't need that reflection hack.

PR #136, opened 2025-04-17
Disable resolution warnings

fmodesto:
## What's changed?
Disable compiler from logging anything during `JavacResoultion`

## What's your motivation?
- https://github.com/PicnicSupermarket/error-prone-support/pull/1644#issuecomment-2809836424

[CompilerMessageSuppressor ](https://github.com/projectlombok/lombok/blob/v1.18.38/src/core/lombok/javac/CompilerMessageSuppressor.java) is from lombok. 
The usage is similar to what lombok does, but in that part codes have differ. 

I have only removed one condition that uses internal lombok code we don't have:
```java
if (listener instanceof CapturingDiagnosticListener) {
    ((CapturingDiagnosticListener) listener).suppress(startPos, endPos);
}
```

@timtebeek 

### Checklist
- [ ] I've used the IntelliJ IDEA auto-formatter on affected files

### DIFF
--- a/src/main/java/org/openrewrite/java/template/internal/CompilerMessageSuppressor.java
+++ b/src/main/java/org/openrewrite/java/template/internal/CompilerMessageSuppressor.java
@@ -1,0 +1,344 @@
+/*
+ * Copyright (C) 2011-2021 The Project Lombok Authors.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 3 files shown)


fmodesto:
Draft due src/test/java/org/openrewrite/java/template/internal/MinimumViableParenthesesTest.java is failing

timtebeek (approved):
Very much appreciated!

PR #138, opened 2025-05-02
Changed processor to inline `before` and `after` templates

JohannisK:
## What's changed?
When generating refaster recipes recipes from remaster template recipes the variable for `before` and `after` are removed and the Objects are instantiated inline.

## What's your motivation?
Fixes: 
- https://github.com/moderneinc/customer-requests/issues/1135


timtebeek:
Merging such that we do not block the work on classpathFromResources, while we figure out the next steps there.

knutwannheden:
Sorry I missed this one. I do have some concerns about performance. What Sam suggested was to lazily initialize the fields.

I think it would be good if we could check the before and after performance of a Refaster recipe with a particularly often used template (I think we have some String recipes). Then we can decide if we need to optimize this.

timtebeek:
Ah I see, thanks for the async review. That subtlety was missed in the example as the fields were left out. Probably best to add that still @JohannisK

knutwannheden:
For the templating use case I think creating the JavaTemplate inline is absolutely idiomatic. Once the execution reaches that point the precondition have typically all been checked. For the template matching as we use it in Refaster recipes, the JavaTemplate itself is the main precondition in the visitor method and I've never really checked how well it would work to create it inline for that use case.

knutwannheden:
Adding a Recipe#getVisitor(ExecutionContext) would have been another interesting idea. But not sure how we feel about that.

PR #139, opened 2025-05-02
Added implementation for classpathFromResource in refasten templates

JohannisK:
## What's changed?
Changed implementation in TemplateCode when generating refaster recipes to use classPathFromResource

## What's your motivation?
Fixes:
- https://github.com/moderneinc/customer-requests/issues/1136

### DIFF
--- a/src/main/java/org/openrewrite/java/template/internal/TemplateCode.java
+++ b/src/main/java/org/openrewrite/java/template/internal/TemplateCode.java
@@ -70,7 +70,11 @@ public class TemplateCode {
                 // See https://github.com/openrewrite/rewrite-templating/issues/86
                 // String classpath = jarNames.stream().map(jarName -> '"' + jarName + '"').sorted().collect(joining(", "));
                 // builder.append("\n    .javaParser(JavaParser.fromJavaVersion().classpath(").append(classpath).append("))");
-                builder.append("\n    .javaParser(JavaParser.fromJavaVersion().classpath(JavaParser.runtimeClasspath()))");
+                // builder.append("\n    .javaParser(JavaParser.fromJavaVersion().classpath(JavaParser.runtimeClasspath()))");
+                builder.append("\n    .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, ");
+                builder.append(jarNames.stream().map(jarName -> '"' + jarName + '"').collect(joining(", ")));
+                builder.append("))\n    ");
+
             }
             return builder.toString();
         } catch (IOException e) {
--- a/src/test/resources/refaster/EscapesRecipes.java
+++ b/src/test/resources/refaster/EscapesRecipes.java
@@ -95,13 +95,13 @@ public class EscapesRecipes extends Recipe {
                     JavaTemplate.Matcher matcher;
                     if ((matcher = JavaTemplate
                             .builder("String.format(\"\\\"%s\\\"\", com.sun.tools.javac.util.Convert.quote(#{value:any(java.lang.String)}))")
-                            .javaParser(JavaParser.fromJavaVersion().classpath(JavaParser.runtimeClasspath()))
+                            .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "tools"))
                             .build().matcher(getCursor())).find()) {
                         maybeRemoveImport("com.sun.tools.javac.util.Convert");
                         return embed(
                                 JavaTemplate
                                         .builder("com.sun.tools.javac.util.Constants.format(#{value:any(java.lang.String)})")
-                                        .javaParser(JavaParser.fromJavaVersion().classpath(JavaParser.runtimeClasspath()))
+                                        .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "tools"))
                                         .build().apply(getCursor(), elem.getCoordinates().replace(), matcher.parameter(0)),
                                 getCursor(),
                                 ctx,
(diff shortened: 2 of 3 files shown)


src/main/java/org/openrewrite/java/template/internal/TemplateCode.java
```diff
@@ -70,7 +70,11 @@ public static <T extends JCTree> String process(T tree, List<JCTree.JCVariableDe
                 // See https://github.com/openrewrite/rewrite-templating/issues/86
                 // String classpath = jarNames.stream().map(jarName -> '"' + jarName + '"').sorted().collect(joining(", "));
                 // builder.append("\n    .javaParser(JavaParser.fromJavaVersion().classpath(").append(classpath).append("))");
-                builder.append("\n    .javaParser(JavaParser.fromJavaVersion().classpath(JavaParser.runtimeClasspath()))");
+                // builder.append("\n    .javaParser(JavaParser.fromJavaVersion().classpath(JavaParser.runtimeClasspath()))");
+                builder.append("\n    .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, ");
+                builder.append(jarNames.stream().map(jarName -> '"' + jarName + '"').collect(joining(", ")));
```
JohannisK, line 75: `jarNames` is constructed from `ClasspathJarNameDetector`. Do we need to check if the resulting list is correct and complete, including dependencies from type tables?

src/test/resources/refaster/EscapesRecipes.java
```diff
@@ -95,13 +95,13 @@ public J visitMethodInvocation(J.MethodInvocation elem, ExecutionContext ctx) {
                     JavaTemplate.Matcher matcher;
                     if ((matcher = JavaTemplate
                             .builder("String.format(\"\\\"%s\\\"\", com.sun.tools.javac.util.Convert.quote(#{value:any(java.lang.String)}))")
-                            .javaParser(JavaParser.fromJavaVersion().classpath(JavaParser.runtimeClasspath()))
+                            .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "tools"))
```
JohannisK, line 98: How do we "know" that tools is in fact on the classpath or in a type table?
How do we handle situations where the names listed do not match the class path/typetables?

JohannisK:
There are several tests for `TemplateProcessor` that fail, because the ExecutionContext is not available there.

src/main/java/org/openrewrite/java/template/internal/TemplateCode.java
```diff
@@ -64,12 +64,11 @@ public static <T extends JCTree> String process(T tree, List<JCTree.JCVariableDe
...
-                // It might be preferable to enumerate exactly the needed dependencies rather than the full classpath
-                // But this is expedient
-                // See https://github.com/openrewrite/rewrite-templating/issues/86
-                // String classpath = jarNames.stream().map(jarName -> '"' + jarName + '"').sorted().collect(joining(", "));
-                // builder.append("\n    .javaParser(JavaParser.fromJavaVersion().classpath(").append(classpath).append("))");
-                builder.append("\n    .javaParser(JavaParser.fromJavaVersion().classpath(JavaParser.runtimeClasspath()))");
+                String joinedJarNames = jarNames.stream().collect(joining(", ", "\"", "\""));
+                builder.append("\n    .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, ")
+                        .append(joinedJarNames)
+                        .append("))\n    ");
```
timtebeek, line 70: Hmm; this now breaks the tests we have for the TemplateProcessor, as the `ctx` can not be found there, and it looks like that might not be easy to introduce where called from openrewrite/rewrite.
https://github.com/openrewrite/rewrite-templating/blob/2c9dcfd3f3186851d871b16531e63430dbc2fb6f/src/test/resources/template/ParameterReuseRecipe.java#L23-L28

Luckily it's not a feature we use often, but still it'd be a shame to break that here.

knutwannheden: We do use it in rewrite-micrometer, IIRC.

timtebeek: There's indeed three places:
1. https://github.com/openrewrite/rewrite-apache/blob/df33d276534aaaea4c2f2f93ad681954084e09b2/src/main/java/org/openrewrite/apache/commons/lang/IsNotEmptyToJdk.java#L74-L77
2. https://github.com/openrewrite/rewrite-migrate-java/blob/6b98403236ad24c7d949750cd1c7af5fb94eebd8/src/main/java/org/openrewrite/java/migrate/UseJavaUtilBase64.java#L126-L127
3. https://github.com/openrewrite/rewrite-micrometer/blob/277c471b5d22db127cd86f00eff7e8ac933c8c9b/src/main/java/org/openrewrite/micrometer/misk/MigrateEmptyLabelMiskCounter.java#L71-L73

timtebeek: I've deferred changing things here for now with a new condtional: https://github.com/openrewrite/rewrite-templating/pull/139/commits/b44be876bf5f62c4dbc5d378d797505ebc9f4170

We _could_ change the `Semantics` class and `PattenrBuilder` to pass along an additional `ExecutionContext` argument, but I figured we need not go through that trouble just yet for those rare cases.

https://github.com/openrewrite/rewrite-templating/blob/022373a047323b6a401d13adc986284f290363ea/src/main/java/org/openrewrite/java/template/Semantics.java#L24-L34

src/main/java/org/openrewrite/java/template/processor/TemplateDescriptor.java
```diff
@@ -148,7 +148,15 @@ public String toJavaTemplateBuilder(int pos) {
...
-        return TemplateCode.process(tree, method.getReturnType().type, method.getParameters(), typeParameters, pos, method.restype.type instanceof Type.JCVoidType, true);
+        return TemplateCode.process(
+                tree,
+                method.getReturnType().type,
+                method.getParameters(),
+                typeParameters,
+                pos,
+                method.restype.type instanceof Type.JCVoidType,
+                true,
+                true);
```
timtebeek, line 159: We could add an option here to make it opt out/in to use classpathFromResources, much like folks can switch between javax/jakarta generated annotations.

knutwannheden (approved):

PR #140, opened 2025-05-03
Support adding static imports for method invocations

timtebeek:
## What's your motivation?
Follow the instructions defined on rules to replace qualified method invocations with static imports.

I've limited this to top level `J.MethodInvocation` only, to cover a common case with as little complexity as possible.
I've also optimistically assumed the goal is to always use static imports when the annotation is used, as per
https://github.com/search?q=repo%3APicnicSupermarket%2Ferror-prone-support+%22%40UseImportPolicy%28%22&type=code

Handling annotation arguments is slightly awkward with assignments, and possible static imports, and not needed for now.
Handling anything beyond direct method invocations comes with a performance penalty as well, that I'd wanted to avoid.

## Any additional context
- Saves us from having `UseStaticImports` twice in https://github.com/openrewrite/rewrite-testing-frameworks/pull/713


timtebeek:
Building on top of this work, and can't imagine there would be heavy criticism of this change, so merging already.

PR #145, opened 2025-05-04
Added support for switching OpenJDK versions in the build process.

fmodesto:
## What's changed?
Added support for switching OpenJDK versions in the build process.

## Motivation
While debugging issues in the error-prone-contrib project, I needed to run the project using Java 17. This change allows flexible JDK version switching during development and testing.

## Usage
Updated the jdkVersion in Gradle settings and refresh the project in IntelliJ IDEA, it will setup itself with the sprcified JDK.

From command-line, default is 8, but we can run with 11, 17 and 21:
```bash
./gradlew -PjdkVersion=11 build
```

---

Do you think this is something worth while pursuing? 

Currently only Java 8 builds successfully.
- Java 8: Builds successfully
- Java 11: Test failures due to generic-related warnings (captured before `CompilerMessageSuppressor` can process them)
- Java 17/21: Test failures from minor code generation differences (affects `Recipe.getDescription()` output)

We could use @RecipeDescriptor for most of the tests, to avoid generating the code in `Recipe.getDescription()`.
And move the tests that cause runtime warnings to new test that doesn't check for this warnings.


timtebeek:
Looks like that could be worthwhile, yes thanks! Maybe even add a matrix build using 8 & 17 to verify the most used targets.

fmodesto:
I have fix the tests to work with openJDK 11, 17 and 21. 

In java 8 and 11 the annotations `@BeforeTemplate()` in the `description()` have parenthesis, that is fix in java 17.

I have added this, but not sure if is better to update all the tests to not generate the code.
```java
                String templateCode = copy.toString().trim()
                        .replace("@BeforeTemplate()", "@BeforeTemplate")
                        .replace("@AfterTemplate()", "@AfterTemplate");
```

Should we allow other versions? currently is limited by "org.openrewrite:rewrite-java-$jdkVersion:latest.integration" but we can add some logic to find the correct dependency

> Looks like that could be worthwhile, yes thanks! Maybe even add a matrix build using 8 & 17 to verify the most used targets.

I'm not familiar with gradle. Most of the changes in `build.gradle.kts` are from ChatGPT.

timtebeek (approved):
Thanks a lot for the help adjusting to various Java versions here @fmodesto !

PR #153, opened 2025-07-22
Allow empty after block for easy element removal

timtebeek:
- As seen on https://github.com/openrewrite/rewrite-testing-frameworks/pull/767/files#r2219877145


src/test/resources/refaster/FindListAddRecipe.java
```diff
@@ -64,7 +64,6 @@ public String getDescription() {
     public TreeVisitor<?, ExecutionContext> getVisitor() {
         JavaVisitor<ExecutionContext> javaVisitor = new AbstractRefasterJavaVisitor() {
             JavaTemplate before;
-            JavaTemplate after;
```
timtebeek, line 67: This one and below never had an `after` template method to begin with; the recipes aim to find only.

PR #157, opened 2025-07-31
Reverting use of static imports for Collections within Refaster examples, as the generated result does not retain the static configuration

steve-aom-elliott:
Fixing the tests


timtebeek (approved):

timtebeek:
Still some small fixes that can be applied to get rid of that last warning. 
https://github.com/openrewrite/rewrite-templating/actions/runs/16657115511/job/47144812447?pr=157#step:12:17

PR #162, opened 2025-08-16
Use Java 21 for tests

timtebeek:


build.gradle.kts
```diff
@@ -70,14 +69,13 @@ nexusPublishing {
 
 java {
     toolchain {
-        languageVersion.set(JavaLanguageVersion.of(jdkVersion))
+        languageVersion.set(of(8))
```
shanman190, line 72: optional: we may want to avoid the static import here so that we can use recipes to modify these if we desire to.

shanman190 (approved):

PR #178, opened 2026-01-29
Add BeforeTemplate precondition to all generated recipes

timtebeek:
## Summary
- Add `Preconditions.not(new UsesType<>("com.google.errorprone.refaster.annotation.BeforeTemplate", true))` as a precondition to every generated recipe
- This prevents recipes from running on files that contain the `@BeforeTemplate` annotation (i.e., the original Refaster template source files)

## Changes
- Add `Precondition.Not` class to `Precondition.java` for wrapping negated preconditions
- Add `NOT_REFASTER_TEMPLATE` constant in `RecipeWriter.java`
- Modify precondition generation to always include this check (combined with `Preconditions.and` when other preconditions exist)
- Update all test expectation files (35 files)

## Test plan
- [x] All existing tests pass (`./gradlew test`)
- [x] Generated recipes include the new precondition structure
- [x] Recipes with existing preconditions are wrapped with `Preconditions.and`
- [x] Recipes without existing preconditions use just the `NOT_REFASTER_TEMPLATE` check

🤖 Generated with [Claude Code](https://claude.com/claude-code)


timtebeek:
Thanks yes figured as much.. Looking to replicate here indeed, as we're now seeing some unfortunate false positives when running our best practices on themselves. :)

